PR #80, opened 2015-03-07
Reactive Streams support for get() operation

### LINKED ISSUE #74: Reactive streams for bambooStorage.get()
We should provide an option for using observable created for get() operation as reactive stream:

For example:

```java
Observable<List<User>> observableUsers = bambooStorage
  .get()
  .listOfObjects(User.class)
  .withMapFunc(User.MAP_FROM_CURSOR)
  .withQuery(Users.QUERY_ALL)
  .prepare()
  .createObservableStream();

observableUsers.subscribe(new Observer<List<Users>>() {
  @Override public void onNext(List<User> users) {
      // will be called after each change in users table
  }

  // ...
});

### PR DESCRIPTION
**Pull request by artem-zinnatullin**
Closes #74

@nikitin-da take a look, please

### DIFF
--- a/bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/operation/get/PreparedGetCursor.java
+++ b/bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/operation/get/PreparedGetCursor.java
@@ -4,12 +4,16 @@ import android.database.Cursor;
 import android.support.annotation.NonNull;
 
 import com.pushtorefresh.android.bamboostorage.BambooStorage;
-import com.pushtorefresh.android.bamboostorage.operation.PreparedOperation;
+import com.pushtorefresh.android.bamboostorage.operation.PreparedOperationWithReactiveStream;
 import com.pushtorefresh.android.bamboostorage.query.Query;
 import com.pushtorefresh.android.bamboostorage.query.RawQuery;
 
+import java.util.HashSet;
+import java.util.Set;
+
 import rx.Observable;
 import rx.Subscriber;
+import rx.functions.Func1;
 
 public class PreparedGetCursor extends PreparedGet<Cursor> {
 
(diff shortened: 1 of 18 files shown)

### CODE REVIEW

File: bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/operation/get/PreparedGetCursor.java [hunk 1]
```diff
@@ -42,6 +46,32 @@
         });
     }
 
+    @NonNull @Override public Observable<Cursor> createObservableStream() {
+        final Set<String> tables;
+
+        if (query != null) {
+            tables = new HashSet<>(1);
```
nikitin-da: Why is capacity one?

artem-zinnatullin: Just memory optimization: Regular `Query` can work with only one table (see next line), while `RawQuery` can work with multiple tables (join and so on)

bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/query/DeleteQuery.java
```diff
@@ -16,4 +18,26 @@ public DeleteQuery(@NonNull String table, @Nullable String where, @Nullable Stri
...
+        if (where != null ? !where.equals(that.where) : that.where != null) return false;
+        if (!Arrays.equals(whereArgs, that.whereArgs)) return false;
+
+        return true;
+    }
+
+    @Override
+    public int hashCode() {
+        int result = table.hashCode();
+        result = 31 * result + (where != null ? where.hashCode() : 0);
```
nikitin-da: Can we add Util class for this?

artem-zinnatullin: Hm, you are talking about `Util` class that we discussed on last conference?

I just used code generation in IDEA

nikitin-da: Yeah, unfortunately no way to use library.
If we will decide to use checkstyle with magic number watchdog, it will be problem =)

Issue #72, opened 2015-03-06
Documentation!


artem-zinnatullin:
There is no JavaDoc at all...

PR #86, opened 2015-03-09
More JavaDoc

artem-zinnatullin:
Some bits of JavaDoc, refs #72 

@nikitin-da PTAL

### DIFF
--- a/bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/BambooStorage.java
+++ b/bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/BambooStorage.java
@@ -20,44 +20,136 @@ import rx.Observable;
 
 public interface BambooStorage {
 
+    /**
+     * Prepares "execute sql" operation for BambooStorage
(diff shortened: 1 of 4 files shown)


bamboo-storage/src/main/java/com/pushtorefresh/android/bamboostorage/BambooStorage.java
```diff
@@ -20,44 +20,136 @@
...
+         * Notifies subscribers about change in set of tables
+         * One operation can affect multiple tables, to reduce number of notifications
+         * you can call this method once and provide set of tables that were changed
+         *
+         * @param affectedTables set of affected tables
+         */
         void notifyAboutChanges(@NonNull Set<String> affectedTables);
 
+        /**
+         * BambooStorage implementation could not provide support for transactions
```
nikitin-da: Do you mean "by default doesn't provide"?

artem-zinnatullin: By default: `BSSQLiteDatabase` provides support for transactions, I was wanted to say that different implementation of `BambooStorage` could not provide support for transactions

PR #131, opened 2015-03-18
Add info about Put operation to README, improvements

artem-zinnatullin:
@nikitin-da PTAL

Btw: you can click on `View` on the file to see how it will look without diffs


nikitin-da:
`shut_up_and_take_my_money.png`

artem-zinnatullin:
Good idea :)

PR #135, opened 2015-03-20
RxJava as optional dependency

### LINKED ISSUE #117: Make RxJava optional dependency

artem-zinnatullin:
Closes #117 

I checked it with `Sample App`, but I don't know how to write proper test for that, it will require a separate module with `provided 'io.reactivex.rxjava` to see crash without it  

@nikitin-da please review

### DIFF
--- a/storio/src/main/java/com/pushtorefresh/storio/util/EnvironmentUtil.java
+++ b/storio/src/main/java/com/pushtorefresh/storio/util/EnvironmentUtil.java
@@ -1,13 +1,16 @@
 package com.pushtorefresh.storio.util;
 
+import android.support.annotation.NonNull;
+
 public class EnvironmentUtil {
 
     /**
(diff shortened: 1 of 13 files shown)


storio/src/main/java/com/pushtorefresh/storio/util/EnvironmentUtil.java
```diff
@@ -0,0 +1,46 @@
...
+    /**
+     * True if RxJava is on classpath, false otherwise
+     */
+    public static final boolean IS_RX_JAVA_AVAILABLE = hasRxJava();
+
+    private EnvironmentUtil() {
+    }
+
+    // thanks Retrofit for that piece of code
+    private static boolean hasRxJava() {
```
nikitin-da: cool!

PR #136, opened 2015-03-20
Add documentation for Operations Builders, throw exceptions from Operations Builders prepare()

### LINKED ISSUE #43: Add checks for all Builder classes

artem-zinnatullin:
Closes #43 

@nikitin-da PTAL

### DIFF
--- a/storio/src/main/java/com/pushtorefresh/storio/db/operation/delete/PreparedDelete.java
+++ b/storio/src/main/java/com/pushtorefresh/storio/db/operation/delete/PreparedDelete.java
@@ -8,7 +8,7 @@ import com.pushtorefresh.storio.db.query.DeleteQuery;
 
 import java.util.Collection;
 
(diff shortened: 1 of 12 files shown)


nikitin-da:
It looks like you don't sleep at all

storio/src/main/java/com/pushtorefresh/storio/db/operation/delete/PreparedDelete.java
```diff
@@ -8,7 +8,7 @@
 
 import java.util.Collection;
 
-public abstract class PreparedDelete<T> implements PreparedOperation<T>{
+public abstract class PreparedDelete<T> implements PreparedOperation<T> {
```
nikitin-da: by the way, does we need common checkstyle config?

artem-zinnatullin: I think yes, but it should be short and non-strict, I'll create an issue

PR #143, opened 2015-03-21
Don't throw exception from notifyAboutChanges() if RxJava is not available

artem-zinnatullin:
Allows to use `StorIODb` without `RxJava` and modify tables.

Case:
1. No `RxJava`
2. User making change in `StorIODb` and Operation tries to call `StorIODb.Internal#notifyAboutChanges()`
3. Crash with "RxJava is not available` exception

Now notifying about changes won't require `RxJava`, but observing changes will.

@nikitin-da PTAL


nikitin-da:
Damn, tests will be helpful. What did you say about separate module without `rx` dependency?

artem-zinnatullin:
Okay,  I'll create it

вс, 22 марта 2015, 1:01, Dmitrii Nikitin <notifications@github.com>:

PR #160, opened 2015-03-25
Draft of PreparedGet for StorIOContentProvider

artem-zinnatullin:
Just a draft, implementation is not full.

@nikitin-da PTAL

after merging I'll continue implementing it


nikitin-da:
super!

Issue #167, opened 2015-03-27
Divide StorIO to several modules


artem-zinnatullin:
* `StorIOCommon`
* `StorIODb`
* `StorIOContentProvider`
* Possible in future: `StorIORealm`

And we will be able to ship `StorIO` as several jars with minimum classes. Not everybody needs db and content provider at same time

ALSO: May be it's better to rename `StorIODb` to `StorIOSQLiteDb`, because `Realm` is db too

nikitin-da:
Hm, not bad.

PR #168, opened 2015-03-28
Add query tests with raw quert

### LINKED ISSUE #67: Integration tests for bambooStorage.get().withQuery(rawQuery)

nikitin-da:
Closes #67 
@artem-zinnatullin Check please.

### DIFF
--- a/storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/BaseTest.java
+++ b/storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/BaseTest.java
@@ -2,6 +2,7 @@ package com.pushtorefresh.storio.db.integration_test.impl;
 
 import android.database.sqlite.SQLiteDatabase;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.support.test.InstrumentationRegistry;
 
 import com.pushtorefresh.storio.db.StorIODb;
@@ -16,6 +17,7 @@ import org.junit.Before;
 import java.util.List;
 
 import static junit.framework.Assert.assertEquals;
+import static junit.framework.Assert.assertNotNull;
 import static junit.framework.Assert.assertTrue;
 
 public abstract class BaseTest {
@@ -39,7 +41,8 @@ public abstract class BaseTest {
                 .executeAsBlocking();
     }
 
-    @NonNull List<User> getAllUsers() {
+    @Nullable
+    List<User> getAllUsers() {
         return storIODb
                 .get()
                 .listOfObjects(User.class)
@@ -51,12 +54,14 @@ public abstract class BaseTest {
                 .executeAsBlocking();
     }
 
-    @NonNull User putUser() {
+    @NonNull
+    User putUser() {
         final User user = TestFactory.newUser();
         return putUser(user);
     }
 
-    @NonNull User putUser(@NonNull final User user) {
+    @NonNull
+    User putUser(@NonNull final User user) {
 
         final PutResult putResult = storIODb
                 .put()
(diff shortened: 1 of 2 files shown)


artem-zinnatullin:
@nikitin-da checkstyle :)

storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/BaseTest.java
```diff
@@ -102,4 +111,19 @@
...
     }
+
+    void checkNotEmptyAndEquals(@Nullable final List<User> first,
+                                @Nullable final List<User> second) {
+        assertNotNull(first);
+        assertNotNull(second);
+        assertEquals(first.size(), second.size());
+
+        for (User userFromFirst : first) {
+            assertTrue(second.contains(userFromFirst));
```
artem-zinnatullin: This check is not correct, it checks that lists contain same items, but order of items can be different. I remember that I wrote own check for comparing lists, but may be better to use `assertEquals(firstList, secondList)`? Standard `List` implementations have correctly overridden `equals()`

storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/QueryTest.java
```diff
@@ -256,4 +269,97 @@
...
+            lengthSum += user.getEmail().length();
+        }
+        final int avrLength = lengthSum / users.size();
+        for (User user : users) {
+            if (user.getEmail().length() > avrLength) {
+                usersWithLongName.add(user);
+            }
+        }
+
+        final String query = "Select * from " + User.TABLE
```
artem-zinnatullin: WOW, `length()`, `avg()` that's why `SQL` beats `NoSQL`

storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/BaseTest.java
artem-zinnatullin: Or you don't want to think about order of items in that check? If so -> better to rename method to something like `checkNotEmptyAndContainsSameElements()` and add comment about order of items :)

storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/QueryTest.java
```diff
@@ -256,4 +269,97 @@
...
+                        .build())
+                .prepare()
+                .executeAsBlocking();
+
+        assertNotNull(usersFromQuery);
+        assertEquals(1, usersFromQuery.size());
+        assertEquals(testUser, usersFromQuery.get(0));
+    }
+
+    @Test public void queryWithRawQuerySqlInjection() {
```
artem-zinnatullin: LOL :+1:

storio/src/androidTest/java/com/pushtorefresh/storio/db/integration_test/impl/BaseTest.java
nikitin-da: Yes, you are right `checkNotEmptyAndContainsSameElements()` will be more clearly.
Because I'm not sure that `Sqllite` saves order by default.

artem-zinnatullin: By default `SQLite` will return results in same order what they were stored in the db

artem-zinnatullin:
@nikitin-da can you please fix it and I'll divide `StorIO` to several modules #167

nikitin-da:
Updated

artem-zinnatullin:
Great :+1:

PR #180, opened 2015-03-30
PreparedPut for StorIOContentProvider with PreparedPutObject impl

artem-zinnatullin:
Plus little refactoring (sorry for that..)

@nikitin-da take a look, please

Part of #164 

I'll add unit tests for `PreparedPut` after finishing all `PreparedPut` variants


nikitin-da:
Cool!!
Annotations from new line produce many changes=)

artem-zinnatullin:
Yeah, I decided to use standard IDEA formatting rules :smile:

nikitin-da:
Ivan is very angry about inline style;)

artem-zinnatullin:
Haha :)

Actually, I don't love inline style non-determenistic, when line is too long, it will wrap annotation, if line is not very long, it won't..

PR #185, opened 2015-03-30
Remove ContentValues argument from DefaultPutResolver.getIdColumnName()

### LINKED ISSUE #178: StorIOSQLiteDb.DefaultPutResolver getTable() should has ContentValues parameter
It can be useful for some libraries/frameworks or user code that analyses `ContentValues` to resolve table name.

For example, user may want to have one `PutResolver` for all classes

artem-zinnatullin:
Closes #178 

I decided to remove `ContentValues` param from `DefaultPutResolver.getIdColumnName()` for simplicity and better performance (no unneeded parameter passing).

In 99% of cases users won't use `ContentValues` to resolve `idColumnName`, it's a work for frameworks and libs based on `StorIO`, so we can keep simpler API for most of our users :)

@nikitin-da PTAL


nikitin-da:
Yeh, it was mistake =(

PR #186, opened 2015-03-30
StorIOContentProvider Query Builders Object... varargs

### LINKED ISSUE #176: Query arguments can be Object...
It's very annoying to write `String.valueOf()` and so for query arguments

Example:

```java
Query query = new Query.Builder()
  .query("SELECT * FROM users WHERE last_activity_timestamp >= ?")
  .args(String.valueOf(System.currentTimemillis())) // annoying, especially when you have several arguments
  .build();
```

We can allow passing `Object... args` and immediately call `object.toString()` to store immutable object's `String` representation

```java
Query query = new Query.Builder()
  .query("SELECT * FROM users WHERE last_activity_timestamp >= ?")
  .args(System.currentTimemillis())
  .build();
```

I think it will be huge improvement to `StorIO` API.

@nikitin-da your opinion?

artem-zinnatullin:
Closes #176 

@nikitin-da please review


nikitin-da:
Nice!

PR #187, opened 2015-03-30
Improve DefaultPutResolver work with id as Object

### LINKED ISSUE #175: DefaultPutResolver should work with id as Object, not as Long
See #173

artem-zinnatullin:
And remove ContentValues param from `getIdColumnName()`

Finally closes #175 

@nikitin-da PTAL

### DIFF
--- a/storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/put/DefaultPutResolver.java
+++ b/storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/put/DefaultPutResolver.java
@@ -26,11 +26,13 @@ public abstract class DefaultPutResolver<T> implements PutResolver<T> {
     /**
      * Provides field name that uses for store internal identifier.
      * You can override this to use your custom name.
+     * <p>
+     * Default value is <code>BaseColumns._ID</code>
      *
      * @return column name to store internal id.
      */
     @NonNull
-    protected String getIdColumnName(@NonNull ContentValues contentValues) {
+    protected String getIdColumnName() {
         return BaseColumns._ID;
     }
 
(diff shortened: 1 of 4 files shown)


storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/put/DefaultPutResolver.java
```diff
@@ -75,14 +82,14 @@ private PutResult insert(@NonNull StorIOSQLiteDb storIOSQLiteDb, @NonNull Conten
...
-                                     @NonNull final String idFieldName,
-                                     @NonNull Object id) {
+                                     @NonNull String idFieldName,
+                                     @NonNull String id) {
 
         final int numberOfUpdatedRows = storIOSQLiteDb.internal().update(
                 new UpdateQuery.Builder()
                         .table(table)
                         .where(idFieldName + "=?")
-                        .whereArgs(String.valueOf(id))
```
artem-zinnatullin: `String.valueOf(object)` will return `String` `"null"` if object will be null, which is not very obvious behaviour for user :)

nikitin-da: NonNull annotation allow to avoid this, isn't it?

artem-zinnatullin: Lol, actually check on line 63 excludes case with `null`, so you are right :+1: 

But I think `idObject` and `idAsString` is more clear, 
If you'll agree, I'll rename `idObject` to `idAsObject`

nikitin-da: Yes, it will be slightly more clearly

artem-zinnatullin: Okay, preparing commit, I'll notify you when PR will be updated

artem-zinnatullin:
@nikitin-da updated

WTF is going on with my Android Studio, it always replacing `<p>` to `</p>` in JavaDoc and vice versa, uh

nikitin-da:
`vice versa` is symbolic

artem-zinnatullin:
Symbolic? Wat? :D I didn't get it

nikitin-da:
`Vice versa` means `the other way around`. 
So the fact that you receive this phrase instead of  tag `</p>` it is some kind of `the other way around`.

artem-zinnatullin:
Oookay..)

PR #207, opened 2015-04-04
Improve StorIOSQLiteDb Delete Result containers

artem-zinnatullin:
@nikitin-da PTAL


storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/delete/DeleteResult.java
```diff
@@ -2,28 +2,81 @@
...
+    private final int numberOfRowsDeleted;
 
-    private DeleteResult(int numberOfDeletedRows, @NonNull Set<String> affectedTables) {
-        this.numberOfDeletedRows = numberOfDeletedRows;
-        this.affectedTables = Collections.unmodifiableSet(affectedTables);
+    @NonNull
+    private final String affectedTable;
+
+    private DeleteResult(int numberOfRowsDeleted, @NonNull String affectedTable) {
+        checkNotNull(affectedTable, "Please specify affected table");
```
nikitin-da: Does this check required with nonNull parameter annotation?

storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/delete/DeleteResults.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @param <T> type of objects
+ */
+public class DeleteResults<T> {
+
+    @NonNull
+    private final Map<T, DeleteResult> results;
+
+    private DeleteResults(@NonNull Map<T, DeleteResult> results) {
+        checkNotNull(results, "Please specify Delete Operation results");
```
nikitin-da: The same question about check

storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/delete/DeleteResults.java
```diff
@@ -0,0 +1,90 @@
...
+     */
+    public boolean wasDeleted(@NonNull T object) {
+        return results.containsKey(object);
+    }
+
+    /**
+     * Checks whether particular object was NOT deleted
+     *
+     * @param object object to check
+     * @return true if object was NOT deleter, false if it was deleted
```
nikitin-da: Is there required `code` tag for `true` value?

storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/delete/DeleteResult.java
artem-zinnatullin: Annotation does not guarantee anything, unfortunately. I am not happy to use `checkNotNull()` because of little performance downgrade, but I think it's better to fail fast than allowing to use object with corrupted state.

storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/delete/DeleteResults.java
artem-zinnatullin: Hm, good catch, yes, I think it's better to use `<code>true</code>` than just `true`, and I see typo in `NOT deleter` should be -> `NOT deleted`

I'll create issues for that

PR #214, opened 2015-04-04
Add Delete By Query for StorIOContentResolver

artem-zinnatullin:
Part of #212 

@nikitin-da PTAL

I think we should count number of tests in `StorIO` when `1.0.0` will be ready, just interesting :)


nikitin-da:
Just number of tests or coverage?

artem-zinnatullin:
Both of them!

вс, 5 апр. 2015, 1:24, Dmitrii Nikitin <notifications@github.com>:

Issue #144, opened 2015-03-21
Create separate module with tests without RxJava


artem-zinnatullin:
Module should exclude 'RxJava' from dependencies and contain tests for `StorIO` features that don't require `RxJava`

artem-zinnatullin:
Damn it, just found problem with creating instance of `StorIOSQLiteDb`.

On line 68, we create anonymous class that extends `rx.functions.Func1` and classLoader throws `NoClassDefFound`

PR #223, opened 2015-04-05
Add separate module for tests without RxJava

artem-zinnatullin:
Part of #144 

@nikitin-da please take a look and thank you for suggestion about creating that module, found one error!


storio-test-without-rxjava/build.gradle
```diff
@@ -0,0 +1,45 @@
...
+
+    testCompile rootProject.ext.junit
+    testCompile rootProject.ext.mockitoCore
+    testCompile project(':storio-test-common')
+
+    androidTestCompile rootProject.ext.testingSupportLib
+}
+
+task checkstyle(type: Checkstyle) {
+    description 'Analyzing project and creating report with bad things in the code'
```
nikitin-da: =)

nikitin-da:
Super!

Issue #218, opened 2015-04-05
Compile checkable required params for all builders


artem-zinnatullin:
I really don't like that our builders can be used incorrectly and if user would forget required param he will see this only in runtime, and it will be crash or the application..

We can make incorrect usage of builders non-compilable, there is workaround for this via interfaces: http://blog.crisp.se/2013/10/09/perlundholm/another-builder-pattern-for-java

@nikitin-da do you agree?

artem-zinnatullin:
Another link: http://blog.crisp.se/2013/10/09/perlundholm/another-builder-pattern-for-java

nikitin-da:
For our case it is just right. Can I take it?

artem-zinnatullin:
Yes, will be great! Can you start with PR for one Builder? I'd like to see how it will be done :)

nikitin-da:
Yep

artem-zinnatullin:
Cool

вс, 5 апр. 2015 г. в 23:55, Dmitrii Nikitin <notifications@github.com>:

nikitin-da:
Did you talk about another link?=) 
It was the same (http://blog.crisp.se/2013/10/09/perlundholm/another-builder-pattern-for-java)

artem-zinnatullin:
Lol, my fault, here is another link: http://www.jayway.com/2012/02/07/builder-pattern-with-a-twist/

Or just Google it: `java builder pattern required parameters`

artem-zinnatullin:
I don't remember, but this problem may also be covered in `Effective Java`

nikitin-da:
Thanks
Hm, I don't remember it in this book

artem-zinnatullin:
Just linking PRs here, @nikitin-da please write something like `refs #218` or `part of #218` in PRs  for that issue, it'll be easier to see, what's already done

Refs: #246, #245, #233, #231, #230, #229, #226

nikitin-da:
ok, thank you

2015-04-09 21:35 GMT+03:00 Artem Zinnatullin <notifications@github.com>:

PR #224, opened 2015-04-06
Compile checkable required params

nikitin-da:
@artem-zinnatullin PTAL
Mandatory builder for PreparedGetListOfObjects

### DIFF
--- a/storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/get/PreparedGetCursor.java
+++ b/storio-sqlitedb/src/main/java/com/pushtorefresh/storio/sqlitedb/operation/get/PreparedGetCursor.java
@@ -78,13 +78,28 @@ public class PreparedGetCursor extends PreparedGet<Cursor> {
         }
     }
 
+    interface CommonBuilder<T> {
+
+        /**
+         * Optional: Specifies {@link GetResolver} for Get Operation
+         * which allows you to customize behavior of Get Operation
+         * <p>
+         * Default value is instance of {@link DefaultGetResolver}
+         *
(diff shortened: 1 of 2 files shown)


nikitin-da:
Part #218

nikitin-da:
I did'n find another builders with that issue.
So there are only two places.

artem-zinnatullin:
Just tried in IDE, it allows me to set only one of mandatory params and I can call `prepare()`

:(

artem-zinnatullin:
Put Operation requires `MapFunc` and PutResolver

You can solve this issue via additional classes, it will also require
special order of methods calls.

For example: `withQuery().withMapFunc()`, when `withQuery()` returns
`non-compete` builder with availability to call `withMapFunc()`  and
`withMapFunc()` returns `CompleteBuilder`

I saw something similar in one of links that I noticed in the issue

пн, 6 апр. 2015, 10:00, Dmitrii Nikitin <notifications@github.com>:

nikitin-da:
Damn, I was concentrated only on logical disjunction and forgot about another required fields.

artem-zinnatullin:
:)

пн, 6 апр. 2015, 10:56, Dmitrii Nikitin <notifications@github.com>:

nikitin-da:
Updated. Do we need be able to set `get resolver` at every step?

artem-zinnatullin:
Great, works like charm!

Yes, I think user should be able to set `GetResolver` at every step, it's one of the benefits of the Builder pattern — order of calls does not matter (except required fields).

I am merging PR, but it does not closes #218 because there are builders for `Queries` and some of them have required fields

Good job :+1:

PR #229, opened 2015-04-07
Builer with required fields for PreparedDeleteObjects

nikitin-da:
@artem-zinnatullin PTAL

### DIFF
--- a/storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/operation/delete/PreparedDeleteObjects.java
+++ b/storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/operation/delete/PreparedDeleteObjects.java
@@ -76,20 +76,22 @@ public class PreparedDeleteObjects<T> extends PreparedDelete<DeleteResults<T>> {
 
     /**
      * Builder for {@link PreparedDeleteObjects}
+     * <p>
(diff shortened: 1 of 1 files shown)


storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/operation/delete/PreparedDeleteObjects.java
```diff
@@ -127,8 +129,42 @@ public Builder(@NonNull StorIOContentResolver storIOContentResolver, @NonNull It
...
+            deleteResolver = builder.deleteResolver;
+            mapFunc = builder.mapFunc;
+        }
+
+        /**
+         * {@inheritDoc}
+         */
+        @NonNull
+        @Override
+        public Builder<T> withDeleteResolver(@NonNull DeleteResolver deleteResolver) {
```
artem-zinnatullin: I think, return type should be `CompleteBuilder<T>`? Because user will have to set `MapFunc` again

And probably, there are same problems in previous PRs, sorry, just thought about it now

nikitin-da:
Thank you, I guess in previous it was ok

artem-zinnatullin:
Great! Keep going!

artem-zinnatullin:
Refs #218

PR #231, opened 2015-04-07
Builer with required fields for PreparedGetListOfObjects

nikitin-da:
@artem-zinnatullin PTAL
As we previously discussed:
1. at first step client should specify query, 
2. then mapFunc,
3. he can prepare.


artem-zinnatullin:
Great!

artem-zinnatullin:
Refs #218

Issue #114, opened 2015-03-14
Change final arrays to unmodifiable lists


artem-zinnatullin:
For immutability of queries for example.

nikitin-da:
wow

artem-zinnatullin:
Oh yeah

сб, 14 марта 2015, 12:39, Dmitrii Nikitin <notifications@github.com>:

PR #243, opened 2015-04-09
StorIOSQLite Get Operation does not require RxJava in ClassPath

artem-zinnatullin:
Part of #114

Now user can actually use `StorIOSQLite.get()` without `RxJava`, fixed `ClassLoader` errors via separate classes that hides `RxJava's` classes

@nikitin-da PTAL


nikitin-da:
Sounds like magic

Issue #213, opened 2015-04-04
All Operation Resolvers should construct result of Operation


artem-zinnatullin:
At the moment, all Operations Resolvers returns something like `int`, etc, but `PutResolver`s return `PutResult`.

So, for more unified code style it will be good to make all Operations Resolvers return `OperationResult`.

Also, in future we will be able to make result of some Operation more complex and it will not affect Operation Resolver return type.

@nikitin-da your opinion?

nikitin-da:
Using special class for `Delete resolver` result will be definitely better. 
As for me in `Get Resolver` case return something else instead of `cursor` will be less obvious.
But for unification I guess it will be good decision.

artem-zinnatullin:
Oh, for GetResolver I wanted to keep Cursor as result type

вс, 5 апр. 2015, 1:34, Dmitrii Nikitin <notifications@github.com>:

PR #250, opened 2015-04-09
Delete Resolver creates result of Delete Operation

artem-zinnatullin:
Part of #213 

@nikitin-da PTAL


nikitin-da:
cool!

artem-zinnatullin:
Oh, I already done same in `StorIO-Content-Provider`, so #213 can be closed :)

PR #259, opened 2015-04-10
Compile safe builder for Query

artem-zinnatullin:
Part of #218 

@nikitin-da PTAL btw what do you think about this PR: I moved all fields except `uri` to `CompleteBuilder`, because it's pretty strange, that you can set other params *before* `uri`, may be we should make all `CompleteBuilders` like that?


nikitin-da:
nice idea!

PR #261, opened 2015-04-10
Move optional fields to CompleteBuilder

artem-zinnatullin:
Part of #218

@nikitin-da PTAL


artem-zinnatullin:
I'll update it, @nikitin-da don't merge please

PR #262, opened 2015-04-10
Fix return type in CompleteBuilder

artem-zinnatullin:
Corrected return type for `CompleteBuilder`

@nikitin-da PTAL


artem-zinnatullin:
Part of #218

nikitin-da:
omg I've missed it

PR #263, opened 2015-04-10
Add CompleteBuilder and overload for setting uri

artem-zinnatullin:
Part of #218 

@nikitin-da PTAL


storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/UpdateQuery.java
```diff
@@ -81,9 +79,34 @@ public int hashCode() {
...
+    /**
+     * Compile-time safe part of {@link Builder}
+     */
+    public static class CompleteBuilder extends Builder {
+
+        private String where;
+        private List<String> whereArgs;
+
+        CompleteBuilder(@NonNull Builder builder) {
+            uri = builder.uri;
```
nikitin-da: If we do smth like this
.uri("usersUri").where("name = 'Ivan'").uri(anotherUsersUri);
Will we lost where clause?

artem-zinnatullin: Yes, my bad, I'll fix it

пт, 10 апр. 2015, 22:13, Dmitrii Nikitin <notifications@github.com>:

artem-zinnatullin:
Updated

PR #265, opened 2015-04-10
Content provider delete query builder

artem-zinnatullin:
Added tests, overload for `uri(String)`

@nikitin-da PTAL

### DIFF
--- a/storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/DeleteQuery.java
+++ b/storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/DeleteQuery.java
@@ -92,6 +92,18 @@ public class DeleteQuery {
             this.uri = uri;
             return new CompleteBuilder(this);
         }
+
+        /**
+         * Required: Specifies uri
+         *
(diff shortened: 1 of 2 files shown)


storio-contentresolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/DeleteQuery.java
```diff
@@ -96,33 +94,15 @@ public CompleteBuilder uri(@NonNull Uri uri) {
...
+         * @param uri full URI to query, including a row ID (if a specific record is requested)
          * @return builder
          */
         @NonNull
-        public Builder whereArgs(@Nullable Object... whereArgs) {
-            this.whereArgs = QueryUtil.varargsToList(whereArgs);
-            return this;
+        public CompleteBuilder uri(@NonNull String uri) {
+            this.uri = Uri.parse(uri);
+            return new CompleteBuilder(this);
```
nikitin-da: Same question, as in previous PR. 
I think we need override `uri` setter or just copy all fields in `CompleteBuilder` constructor.

artem-zinnatullin:
Updated

nikitin-da:
:+1:

PR #272, opened 2015-04-10
StorIOSQLite Query builders and tests

artem-zinnatullin:
Compile-time safe builders for `StorIOSQLite` queries

@nikitin-da PTAL btw you used `Query` builder incorrectly in integration tests! Ha!


artem-zinnatullin:
Part of #218

storio-common/src/main/java/com/pushtorefresh/storio/util/Checks.java
```diff
@@ -6,7 +6,7 @@
 public final class Checks {
 
     private Checks() {
-
+        throw new IllegalStateException("No instances please");
```
nikitin-da: Wow is it possible?

PR #280, opened 2015-04-13
Better structure of README, new info about object mapping

artem-zinnatullin:
I duplicated documentation and links to sample and design tests BEFORE examples and AFTER examples, so it'll be easier to find them for new users after reading examples and for old users when they just opened GitHub repository page

@nikitin-da PTAL


nikitin-da:
:+1:

PR #300, opened 2015-04-25
Allow extending PreparedGroupOperation

artem-zinnatullin:
It'll allow users to create their own Group Operations.

Also, later we will provide implementation of group operation for `StorIOSQLite` with transaction support

@nikitin-da PTAL


nikitin-da:
:+1:

PR #301, opened 2015-04-25
Operation resolvers now abstract classes

### LINKED ISSUE #283: Operation Resolvers should be abstract classes

artem-zinnatullin:
Closes #283 

All Operation Resolvers now abstract classes

@nikitin-da PTAL


artem-zinnatullin:
retest this please

PR #302, opened 2015-04-25
Corrections for all builders

### LINKED ISSUE #218: Compile checkable required params for all builders

artem-zinnatullin:
Closes #218

@nikitin-da review please :)


storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/Query.java
```diff
@@ -103,9 +102,7 @@ public String toString() {
     /**
      * Builder for {@link Query}
      */
-    public static class Builder {
-
-        Uri uri;
+    public static final class Builder {
```
nikitin-da: Why did you decide to make it final? Can it be useful for user to extend it for some custom behavior. For example receive `String` instead of `Uri`?

nikitin-da:
:+1:

storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/Query.java
artem-zinnatullin: But how he would apply his custom class?

вс, 26 апр. 2015, 14:54, Dmitrii Nikitin <notifications@github.com>:

nikitin-da: `class CustomBuilder exteds Query.Builder {
   CompleteBuilder uri(java.net.URI javaUri) {
      return(javaUri.toString());
   }
}`
and then
storIOContentResolver()
                .get()
                .cursor()
                .withQuery(new Query.CustomBuilder()
                        .uri(existJavaUri)
                        .build())
`
Of cause, it is far-fetched example, but are you sure that it will be not necessary or just want to hide all except really important?

artem-zinnatullin: It won't work, you'll have to use custom Query class for that or modify
sources of Query class and build StorIO from sources. Yep, I want to hide
and make final everything to be sure that won't have to support inheritance
of our classes

вс, 26 апр. 2015, 19:57, Dmitrii Nikitin <notifications@github.com>:

PR #305, opened 2015-04-27
Finally, final classes

### LINKED ISSUE #227: Hide and make everything private
We are pretty close to release `1.0.0` and many things should be private and/or final: classes, constructors and so on. We can not offer support to all classes in terms of inheritance, may be later we will make some of them extendable again.

artem-zinnatullin:
Many classes made `final`, added javadoc about "internal usage"

Closes #227

@nikitin-da PTAL


nikitin-da:
:+1:

PR #307, opened 2015-04-28
Switch to Android Gradle Plugin 1.2.1

artem-zinnatullin:
@nikitin-da this also fixes problem with Android Studio


nikitin-da:
Yeah!

PR #323, opened 2015-05-11
Use Thread.yield() instead of sleep()

### LINKED ISSUE #320: Use Thread.yield() instead of sleep() in tests
Here: https://github.com/pushtorefresh/storio/blob/a6015e284efc3f1f2238849fbabd881658cd4b1f/storio-sqlite/src/androidTest/java/com/pushtorefresh/storio/sqlite/impl/NotifyAboutChangesTest.java#L42

And here: https://github.com/pushtorefresh/storio/blob/a6015e284efc3f1f2238849fbabd881658cd4b1f/storio-sqlite/src/androidTest/java/com/pushtorefresh/storio/sqlite/impl/NotifyAboutChangesTest.java#L94

artem-zinnatullin:
Closes #320 

@nikitin-da PTAL


nikitin-da:
:+1:

PR #336, opened 2015-05-16
Improve StorIOContentResolver Javadocs

artem-zinnatullin:
Huge pack of improvements to `StorIOContentResolver` Javadocs.

1. Dots at the end of each sentence.
2. Links and formatting.

@nikitin-da PTAL

PR #337, opened 2015-05-17
Fix for thread-safety of notifications system in StorIOSQLite

artem-zinnatullin:
Previous implementation had problem with thread-safety, it could publish one notification several times :(

This patch fixes that, also I've added two new integration tests for notifications system of StorIOSQLite.

@nikitin-da PTAL

It should help with #336 problem and flaky tests.


artem-zinnatullin:
@nikitin-da we can merge this PR first and I'll rebase it on other PRs to check that tests won't fail on them too

nikitin-da:
Great!

PR #347, opened 2015-05-20
RawQuery affectsTables and observesTables

### LINKED ISSUE #342: Add observesTables to the RawQuery
`RawQuery` can be used not only to affect db, but also for querying:

```java
storIOSQLite
  .get()
  .cursor()
  .withQuery(new RawQuery.Builder()
    .query("SELECT * FROM users JOIN tweets ON tweet.user_id = users.user_id WHERE tweet.timestamp > ?")
    .args(System.currentTimeMillis() - 86400)
    .affectedTables("users", "tweets") // this query does not affect db!
    .build())
  .prepare()
  .createObservable()
  .subscribeOn(Schedulers.io())
  .observeOn(AndroidSchedulers.mainThread())
  .subscribe(cursor -> adapter.setCursor(cursor)); // will be updated if users or tweets will be changed
```

I think `tables` instead of `affectedTables` should be good.

@nikitin-da what do you think?

artem-zinnatullin:
Closes #342 

Now `RawQuery` looks good!

@nikitin-da PTAL

### DIFF
--- a/storio-common/src/main/java/com/pushtorefresh/storio/internal/Queries.java
+++ b/storio-common/src/main/java/com/pushtorefresh/storio/internal/Queries.java
@@ -5,9 +5,10 @@ import android.support.annotation.Nullable;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
+import java.util.Set;
 
 /**
- * Util methods for queries
+ * Util methods for queries.
  * <p/>
  * For internal usage only!
  */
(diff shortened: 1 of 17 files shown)


storio-common/src/main/java/com/pushtorefresh/storio/internal/Queries.java
```diff
@@ -67,4 +82,6 @@ private Queries() {
                 ? null
                 : list.toArray(new String[list.size()]);
     }
+
+
```
nikitin-da: Are that empty lines for better formating?

storio-common/src/test/java/com/pushtorefresh/storio/internal/QueriesTest.java
```diff
@@ -77,8 +77,8 @@ public void nonEmptyListToUnmodifiable() {
 
     @SuppressWarnings("ConstantConditions")
     @Test
-    public void listToUnmodifiableIsReallyUnmodifiable() {
-        List<String> unmodifiableList = Queries.listToUnmodifiable(Arrays.asList("1", "2", "3"));
+    public void listToUnmodifiableNullableListIsReallyUnmodifiable() {
```
nikitin-da: =)

nikitin-da:
Good!

storio-common/src/main/java/com/pushtorefresh/storio/internal/Queries.java
artem-zinnatullin: I don't know :D

PR #362, opened 2015-06-04
No more nulls in queries

### LINKED ISSUE #346: Kill nulls in Queries!
Nulls everywhere in StorIO, that's good for performance, but not safe for users, I think that it's better to lose 1% of performance, but provide safety instead.

artem-zinnatullin:
Closes #346.

@nikitin-da PTAL!

### DIFF
--- a/storio-common/src/test/java/com/pushtorefresh/storio/internal/QueriesTest.java
+++ b/storio-common/src/test/java/com/pushtorefresh/storio/internal/QueriesTest.java
@@ -4,69 +4,111 @@ import org.junit.Test;
 
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.HashSet;
 import java.util.List;
+import java.util.Set;
 
+import static java.util.Collections.emptyList;
+import static java.util.Collections.emptySet;
 import static junit.framework.Assert.assertEquals;
 import static junit.framework.Assert.assertNotNull;
 import static junit.framework.Assert.assertTrue;
 
 public class QueriesTest {
 
+    //region Tests for Queries.unmodifiableNonNullListOfStrings()
+
     @SuppressWarnings("ConstantConditions")
     @Test
-    public void nullVarargsToList() {
-        Object[] varargs = null;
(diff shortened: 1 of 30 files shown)


storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/query/DeleteQuery.java
```diff
@@ -187,6 +188,10 @@ public CompleteBuilder whereArgs(@Nullable Object... whereArgs) {
          */
         @NonNull
         public DeleteQuery build() {
+            if (where == null && whereArgs != null && !whereArgs.isEmpty()) {
```
nikitin-da: whereArgs is never null, isn't it?

nikitin-da: No, all right

storio-common/src/test/java/com/pushtorefresh/storio/internal/QueriesTest.java
```diff
@@ -110,4 +210,44 @@ public void nonEmptyListOfStringsToArray() {
...
+        assertEquals("123", Queries.nonNullString("123"));
+    }
+
+    //endregion
+
+    //region Tests for Queries.nullableString()
+
+    @Test
+    public void nullableStringFromNull() {
+        assertEquals(null, Queries.nullableString(null));
```
nikitin-da: Maybe just assertNull?

nikitin-da:
Great!

storio-common/src/test/java/com/pushtorefresh/storio/internal/QueriesTest.java
artem-zinnatullin: Wow, sure :)

I am thinking about AssertJ instead of JUnit, so it'll be replaced with
assertThat(someThing).isNull()
5 июня 2015 г. 1:59 AM пользователь "Dmitrii Nikitin" <

Issue #372, opened 2015-06-08
Add example with AutoValue/AutoParcel entity


artem-zinnatullin:

artem-zinnatullin:
Closed by #374.

PR #374, opened 2015-06-08
Type mapping now works for subclasses

### LINKED ISSUE #365: Type mapping with AutoValue/AutoParcel
Link to discussion on reddit: https://www.reddit.com/r/androiddev/comments/38nx2t/weekly_anything_goes_thread/crxm900

The problem:

`SQLiteTypeMapping` and `ContentResolverTypeMapping` store pairs for type mapping with exact class, for example: `[(Tweet.class, TweetPutResolver), (Tweet.class, TweetGetResolver)...]`.

But with `AutoValue` and `AutoParcel` it won't resolve required class.

> There was one hiccup with AutoParcel and SQLiteTypeMapping. AutoParcel generates child class for every annotated class, but it's package private and can't be used when creating type mapping but type mapping created using parent class doesn't work with child object since prepare operation uses getClass to find appropriate type mapping. For example:

```java
@AutoParcel
public abstract class Foo {
    abstract int id();
    abstract String bar();
    public static Foo create(int id, String bar) {
        return new AutoParcel_Foo(id, name);
}

StorIOSQLite storIOSQLite = new DefaultStorIOSQLite.Builder()
.sqliteOpenHelper(helper)
.addTypeMapping(Foo.class, new SQLiteTypeMapping.Builder<Foo>()
    .putResolver(new FooPutResolver())
    .getResolver(new FooGetResolver())
    .deleteResolver(new FooDeleteResolver())
    .build())
.build();

Foo foo = Foo.create(0, "baz"); // foo.getClass returns AutoValue_Foo

storIOSQLite
  .put()
  .object(foo)
  .prepare() // prepare can't find mapping for AutoValue_Foo
  .executeAsBlocking() 
```

> On user side it can be solved by specifying concrete right resolver for every operation or using generated class type in type mapping and resolver declaration neither of which very flexible. How it can be solved in library I have no idea. So for now I stopped using AutoValue for db entity objects (same with Gson based entities since AutoValue doesn't play nice with it either).

artem-zinnatullin:
Closes #365!

@nikitin-da PTAL, see #365 for description of implementation, I think that my implementation is pretty fast and "reflection" used only once, then it's cached.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ Currently in development.
 * API for Humans: Type Safety, Immutability & Thread-Safety
 * Convenient builders with compile-time guarantees for required params. Forget about 6-7 `null` in queries
 * Optional Type-Safe Object Mapping, if you don't want to work with `Cursor` and `ContentValues` you don't have to
-* No reflection and no annotations in core, also `StorIO` is not ORM
+* No reflection in Operations and no annotations in the core, also `StorIO` is not ORM
 * Every Operation over `StorIO` can be executed as blocking call or as `rx.Observable`
 * `RxJava` as first class citizen, but it's not required dependency!
 * `rx.Observable` from `Get` Operation **can observe changes** in `StorIO` and receive updates automatically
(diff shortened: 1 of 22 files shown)


artem-zinnatullin:
Don't merge until I'll add support for better inheritance processing.

artem-zinnatullin:
@nikitin-da I've reimplemented typeMapping resolving, now it's faster and also support deep inheritance and complex relations like in your example with AutoValue, PTAL carefully at whole PR.

Also, I've added unit and integration tests for AutoValue/AutoParcel support.

artem-zinnatullin:
BTW, it also closes #372!

nikitin-da:
Awesome! Infinitely better autoValue support then Gson =)

artem-zinnatullin:
Haha, yeah :)
9 июня 2015 г. 1:00 PM пользователь "Dmitrii Nikitin" <

PR #377, opened 2015-06-09
Better dependencies declaration

artem-zinnatullin:
Little nice thing.

@nikitin-da PTAL

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,35 +34,46 @@ ext {
     buildToolsVersion = '22.0.1'
     compileSdkVersion = 22
     minSdkVersion = 14
-
-    supportAnnotations = 'com.android.support:support-annotations:22.2.0'
-    rxJava = 'io.reactivex:rxjava:1.0.11'
-
-    dagger = 'com.google.dagger:dagger:2.0'
-    daggerCompiler = 'com.google.dagger:dagger-compiler:2.0'
-
-    javaxInject = 'javax.inject:javax.inject:1'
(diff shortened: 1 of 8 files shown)


build.gradle
```diff
@@ -34,36 +34,47 @@ ext {
...
 }
 
+ext.libraries = [
+        // Core libraries
+        storIOCommon                   : project(':storio-common'),
+        supportAnnotations             : 'com.android.support:support-annotations:22.2.0',
+        rxJava                         : 'io.reactivex:rxjava:1.0.11',
+
+        // Libraries for tests and sample app
+        storIOSQLite                   : project(':storio-sqlite'),
```
nikitin-da: Are `storio-sqlite` and etc libraries for test or for sample app?

nikitin-da: I think it's core too

artem-zinnatullin: Core libraries — libraries that needed for `StorIOSQLite` and `StorIOContentResolver` :)

So, I think `StorIOSQLite` and `StorIOContentResolver` are not core libraries.

nikitin-da: Yes, call that modules as `core` is not entirely accurate. But I don't agree that it's just for test or for sample app =)

artem-zinnatullin: Hehe :) So, how should we call them?

nikitin-da: I all these modules in block, maybe `facilities`?

artem-zinnatullin: Noooo :) Sounds bad

I think we can make block `// StorIO parts`

nikitin-da: :+1:

PR #379, opened 2015-06-09
Better declaration of StorIO parts

artem-zinnatullin:
@nikitin-da PTAL


nikitin-da:
looks like a charm

artem-zinnatullin:
:shipit:

PR #387, opened 2015-06-15
Create builders via static methods

### LINKED ISSUE #382: Create builders via static methods
IntelliJ IDEA's suggestions engine doesn't handle `new Something.Builder()` properly, it always suggests me `new Something()` instead of builder even if constructor of `Something` is `private`.

So, I think we should switch to `Something.builder()`, it also requires less symbols in compare to `new Something.Builder()`.

nikitin-da:
@artem-zinnatullin  PTAL, please.
Closes #382


storio-content-resolver/src/androidTest/java/com/pushtorefresh/storio/content_resolver/impl/BaseTest.java
```diff
@@ -30,14 +30,14 @@ public BaseTest() {
     public void setUp() throws Exception {
         super.setUp();
 
-        storIOContentResolver = new DefaultStorIOContentResolver.Builder()
+        storIOContentResolver = DefaultStorIOContentResolver.builder()
```
artem-zinnatullin: Oh yes!

artem-zinnatullin:
:+1:

PR #394, opened 2015-06-16
Improve Delete collection of objects codebase

artem-zinnatullin:
Just little improvement for Delete collection of objects.

@nikitin-da PTAL


nikitin-da:
:+1:

PR #395, opened 2015-06-17
Fix emission checker behavior in case of unexpected value

artem-zinnatullin:
Before:

If value that was received by emission checker was incorrect -> it'll throw exception which could be handled by Observable's subscriber and it brakes internal checks of emission checker, it may or may not handle it as timeout error.

After:

Now I handle all exceptions manually and threat them as Emission Checker's problem, so Observable won't handle it for us.

And finally, it may help with our flaky tests on CI problem.

+ Bonus -> `build time -= 2 minutes`!

### DIFF
--- a/storio-test-common/src/main/java/com/pushtorefresh/storio/test/AbstractEmissionChecker.java
+++ b/storio-test-common/src/main/java/com/pushtorefresh/storio/test/AbstractEmissionChecker.java
@@ -5,6 +5,7 @@ import android.support.annotation.NonNull;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 
 import rx.Subscription;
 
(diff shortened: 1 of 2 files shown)


nikitin-da:
Hm `shouldReceiveChangeAfterEachUpdate` failed=(

artem-zinnatullin:
Uh… I hope, we will see Genymotion for CI soon, *ing slow Android emulator.

nikitin-da:
=) 
There are problems again.
Shell I merge it?

artem-zinnatullin:
Passed!

nikitin-da:
Yeah)

nikitin-da:
Have you change something?

artem-zinnatullin:
No, just relaunched build on Travis 3-4 times… I am thinking about Jenkins or TeamCity setup with powerful slaves hosted on AWS EC2 .

artem-zinnatullin:
Or, we can switch to Robolectric, but as you know, I am not fan of it…

nikitin-da:
Yes, I know, but why?

artem-zinnatullin:
Oh, with Robolectric we will open whole new world of dealing with Robolectric's bugs.

Issue #371, opened 2015-06-08
Put and Delete objects should not require specifying type


artem-zinnatullin:
I think, it's my mistake in the API.

artem-zinnatullin:
@nikitin-da please help me decide:

If we won't require specifying class for `put.objects()` and `delete.objects()` we have to get class of the objects for type mapping.

I see three options:

First option — look through the collection of objects in `prepare()` and create `List<TypeMappingAndObject>` and pass it to the operation in `prepare()`, it'll work in calling thread which will be Main Thread (bad) in most of the cases, but it can fail early if some object does not have type mapping and db won't be affected at all.

Second option — get class of each object when operation executes, it'll be good if operation is async and runs on bg thread (with RxJava, for example), but because we will collect this information for each item during the execution, db could be left in affected state, for example: 0 — okay, 1 — okay, 2 — no type mapping -> exception -> 0 and 1 in db.

Third option — mix of first and second: collect all types mappings in execution (can be async) but before affecting the db.

All of these options will be little-bit slower than current version, but, they make API cleaner and more flexible: you can add objects of different types in the collection and perform put or delete in transaction with just a few lines of code, which is very cool and great.

artem-zinnatullin:
Addition: other operations will throw exception if type mapping is not found in `prepare()` stage, with second and third options we can't do it.

nikitin-da:
First looks good given the availability of the cache, whether the first search take more than 16 ms?
I dislike second way, because I think that it will be worse case if we will left database in unpredictable state.
Third will be more complicated to implement, but does it have other minuses?

artem-zinnatullin:
16ms is not a problem, for us — it's a problem if it'll take >= 0.5 ms I guess, because Android Framework can take about ~5ms on it's own things, GC can take other 5ms and developer should use time left very efficiently…

Also, for huge collections it would take even more time and memory.

I like second option because it's memory allocations free, but I dislike possibility of affected db very much. But…but by default `Put(objects)` and `Delete(objects)` will use transaction in case of SQLite, but for `ContentResolver` they wouldn't, of course.

Also, developer may pass `Iterable` that would load items from network or so on and we will read it until the end in the calling thread… So I guess, we should do it in execution step. Actually, in case of `Iterable`  the only good solution is — second solution, because Iterable can be **really** long.

artem-zinnatullin:
Done!

nikitin-da:
:+1:

PR #397, opened 2015-06-17
Dynamic type mapping for Delete collection of objects

artem-zinnatullin:
Part of #371.

@nikitin-da PTAL


nikitin-da:
Success =)

Issue #396, opened 2015-06-17
Get/Put/Delete should resolve type mapping in "execution" phase


artem-zinnatullin:
If no type mapping was found for Put/Delete of one object -> throw error in execute phase, not in prepare, because rx.Observer should be able to handle this.

artem-zinnatullin:
It's done for Put/Delete Operations, but I think that we should also resolve type mapping for Get Operation in "execute" phase.

PR #399, opened 2015-06-18
Dynamic type mapping for Put object/objecs

artem-zinnatullin:
Near to end of #371.

@nikitin-da PTAL!


artem-zinnatullin:
Also, it's part of #396.

nikitin-da:
Cool!

PR #404, opened 2015-06-18
Put object type mapping error handling

artem-zinnatullin:
Final part of #396!

@nikitin-da PTAL

And sorry for such amount of PRs, I just want to release StorIO :shipit:


nikitin-da:
404 :)

storio-sqlite/src/test/java/com/pushtorefresh/storio/sqlite/operation/put/PutObjectsStub.java
```diff
@@ -106,18 +130,37 @@ static PutObjectsStub newPutStubForMultipleObjectsWithTypeMappingWithoutTransact
     }
 
     void verifyBehaviorForMultipleObjects(@NonNull PutResults<TestItem> putResults) {
+        // should be called once because of Performance!
+        verify(storIOSQLite).internal();
```
nikitin-da: Does the same question actual here?

artem-zinnatullin: I've used `verifyNoMoreInteractions()` so, it's okay to don't have `verify(..., never())` in this case.

Issue #353, opened 2015-05-31
Documentation: Add example of getting result once via Get Operation as rx.Observable


artem-zinnatullin:

nikitin-da:
Is it actual?

artem-zinnatullin:
Yes!
21 июня 2015 г. 10:06 PM пользователь "Dmitrii Nikitin" <
notifications@github.com> написал:

artem-zinnatullin:
@nikitin-da if you have time, I'll be happy to review your PR for this issue

nikitin-da:
Ok, I'll do it this evening.

artem-zinnatullin:
Done!

PR #414, opened 2015-06-22
Better javadoc formatting, but still not perfect

artem-zinnatullin:
Uh, spent two hours on fighting with Gradle + Android Gradle Plugin + Javadoc generating, almost everything is done except cross-module dependencies: for example, for `storio-sqlite` and `storio-content-resolver` javadoc tool does not see classes from `storio-common`…

Anyway, this is better than nothing.

@nikitin-da PTAL and what about release tomorrow (Tuesday)?


nikitin-da:
Sounds great! What can I do for this?
Can I make #353 or something else?

PR #418, opened 2015-06-23
Example of getting one result with Rx from sqlite

nikitin-da:
Part of #353.
@artem-zinnatullin Did you mean something like this?

PR #419, opened 2015-06-23
Example of getting one result with Rx from contentResolver

nikitin-da:
Second part of #353.
@artem-zinnatullin PTAL =)


artem-zinnatullin:
Same comments as for #418.

nikitin-da:
@artem-zinnatullin  Thank you very much for advances! Updated.

artem-zinnatullin:
Thanks!

PR #422, opened 2015-06-24
Fix race condition in integration tests for StorIOContentResolver

artem-zinnatullin:
This PR fixes race condition in integration tests for `StorIOContentResolver`, probably this race condition caused troubles with emulator too.

Description:

Put into `ContentProvider` causes `Changes`, all tests inserts some data in the beginning and we didn't `await` until these inserts cause `Changes` on `RxBus`, so if changes arrive with some delay (on slow build machine, for example), it can break further checks in some tests.

Now I made `await` for changes after inserts.

@nikitin-da PTAL, you can try to re-launch this PR on Travis several times to check that it builds normally without race conditions.

### DIFF
--- a/storio-content-resolver/src/test/java/com/pushtorefresh/storio/contentresolver/integration/IntegrationTest.java
+++ b/storio-content-resolver/src/test/java/com/pushtorefresh/storio/contentresolver/integration/IntegrationTest.java
@@ -4,6 +4,7 @@ import android.content.ContentResolver;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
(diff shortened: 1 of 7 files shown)


nikitin-da:
`shouldReceiveChangeAfterEachInsert` failed =(

artem-zinnatullin:
Damn

artem-zinnatullin:
3 runs on Travis, 1 fail.
On local machine I never get fails, 15+ runs.

nikitin-da:
Anyway now it is a way more stable. Will I merge?

artem-zinnatullin:
5 runs, 1 failure, let's merge this, I'll investigate this problem later.

PR #423, opened 2015-06-25
Fixed wasUpdated

krpiotrek:
Simple fix in case of such code : 
```
final int update = storIOSQLite
	.internal()
	.update(new UpdateQuery.Builder()
	.table(TABLE)
	.build(), cv);
return PutResult.newUpdateResult(update, TABLE);
```

since update() returns an int it's better to check > 0 on library side than client's

### DIFF
--- a/storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/put/PutResult.java
+++ b/storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/put/PutResult.java
@@ -78,7 +78,7 @@ public final class PutResult {
      * {@code false} otherwise.
      */
     public boolean wasUpdated() {
-        return numberOfRowsUpdated != null;
+        return numberOfRowsUpdated != null && numberOfRowsUpdated > 0;
     }
 
     /**
--- a/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/put/PutResult.java
+++ b/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/put/PutResult.java
@@ -106,7 +106,7 @@ public class PutResult {
      * {@link com.pushtorefresh.storio.sqlite.StorIOSQLite}, {@code false} otherwise.
      */
     public boolean wasUpdated() {
-        return numberOfRowsUpdated != null;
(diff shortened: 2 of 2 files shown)


nikitin-da:
Nice fix :+1: 
Thank you very much!

artem-zinnatullin:
Hm, changes looks good to me, but what if we want to represent result of update and number of rows updated `== 0`?

Also, we should cover this with tests!

artem-zinnatullin:
I re-thinked this and I'd agree, that result of Put Operation should be Update (with number of rows updated `>= 1`) or Insert, so the only thing left — tests, I'll submit PR with them soon.

artem-zinnatullin:
I think I have better idea, let's throw Exception from newUpdateResult if number of rows updated <1

nikitin-da:
I think it's `putresolver`'s problem or maybe we should use single-common static creator for insert and update. Otherwise it's not obvious why update result with `0` updated rows is invalid.

artem-zinnatullin:
Update result with `0` updated rows is invalid in terms of Put Operation, because if you've updated `0` rows -> please, make insert, m?

nikitin-da:
What if user'll deside to make custom put resolver with checking if rows was really updated. He will query target rows from db, then if their alredy exist he'll check for equals. So it'll be useful for him to have `putResult` with `0` rows.

nikitin-da:
I want to say that imho it is not quite right to prohibit result with `0` relying on knowledge that encapsulated in another class.

artem-zinnatullin:
Hm, not sure I understood your case:

>What if user'll deside to make custom put resolver with checking if rows was really updated. He will query target rows from db, then if their alredy exist he'll check for equals. So it'll be useful for him to have `putResult` with `0` rows.

This is valid point:

> I want to say that imho it is not quite right to prohibit result with 0 relying on knowledge that encapsulated in another class.

But I strongly dislike that now we can not understand what is the result of Put Operation, previously you can easily check it was insert or update:

`putResult.wasInserted() // true if row was inserted`
`putResult.wasUpdated() // true if one or more rows were updated`

Now to handle case where 0 rows were updated we have to check that `wasInserted() == false && wasUpdated() == false && numberOfRowsUpdated == 0`.

PR #424, opened 2015-06-28
Reduce tests flakiness via awaiting changes after modifying operations on ContentProvider

artem-zinnatullin:
This PR should reduce frequency of failing builds on CI, now we waiting for `Changes` on the `Bus` in all tests that modify the `ContentProvider`.

@nikitin-da PTAL!

I'll relaunch CI several times to see if it fails.


nikitin-da:
Success)

nikitin-da:
Stable!

PR #425, opened 2015-06-30
Consistent behavior of PutResult.newUpdateResult() in case of number …

artem-zinnatullin:
…of rows updated < 1, tests

This PR makes behavior of `PutResult.newUpdateResult()` consistent with `SQLiteDatabase`, if number of rows updated < 1 -> throw exception.

Also, I've noticed that we didn't have tests on `PutResult`! So I've added them.

@nikitin-da PTAL!


artem-zinnatullin:
@nikitin-da hope you'll say LGTM, I am merging this and preparing release 1.0.0!

nikitin-da:
Good=)

PR #456, opened 2015-07-22
Correct info about Rx support

artem-zinnatullin:
Just a little improvement for the README.


artem-zinnatullin:
@nikitin-da PTAL!

PR #459, opened 2015-07-25
80% code coverage for all Android modules

### LINKED ISSUE #450: Add test coverage tool to the build
Also, it should fail the build if coverage is less than some value (80% for example).

artem-zinnatullin:
Resolves #450.

@nikitin-da PTAL!

I am very sorry for such huge PR, but code coverage touched every module and required additions to our test classes in storio-test-common.

I can try to divide it into separate PRs, but it would take few hours/days.


artem-zinnatullin:
Oh shi, 3000+ lines added, I should divide it to separate PRs.

PR #461, opened 2015-07-25
80% code coverage for StorIO-Common

artem-zinnatullin:
Part of #450.

This PR adds tests for StorIO-Common, now coverage of this module is about 99%.

@nikitin-da PTAL


artem-zinnatullin:
retest this please

artem-zinnatullin:
YES, I've finally fixed the problem with test coverage of `OnSubscribeExecuteAsBlocking`.

It was race condition which never happens on build machine with more than one Processor Core but on Travis CI and on our Jenkins we have only one core and test execution became different.

@nikitin-da PTAL!

storio-common/src/test/java/com/pushtorefresh/storio/internal/EnvironmentTest.java
```diff
@@ -0,0 +1,67 @@
...
+            assertEquals("yolo requires RxJava in classpath," +
+                            " please add it as compile dependency to the application",
+                    expected.getMessage()
+            );
+        } finally {
+            // Return previous value of the field
+            field.set(null, prevValue);
+
+            // Restoring FINAL modifier (for better tests performance)
+            modifiersFieldOfTheField.setInt(field, field.getModifiers() & FINAL);
```
nikitin-da: Why should it be faster with`final`? As for me it's not uniquely (http://stackoverflow.com/a/4279442)

nikitin-da:
Well done!

storio-common/src/test/java/com/pushtorefresh/storio/internal/EnvironmentTest.java
artem-zinnatullin: Hm, I think it will be faster because of JIT.
27 июля 2015 г. 8:59 AM пользователь "Dmitrii Nikitin" <

Issue #481, opened 2015-08-05
Deadlock accessing the database - don't use singleton StorIO in multiple threads


tadas-subonis:
So after some time I was able to track the issue to these components:

DefaultStoreIOSQLite.java
```java
   @Override
        public void beginTransaction() {
            synchronized (lock) {
                sqLiteOpenHelper
                        .getWritableDatabase()
                        .beginTransaction();

                numberOfRunningTransactions++;
            }
        }
```

and 

SQLiteConnectionPool.java with SQLiteDatabase and SQLiteSession
```java
    public SQLiteConnection acquireConnection(String sql, int connectionFlags,
            CancellationSignal cancellationSignal) {
        return waitForConnection(sql, connectionFlags, cancellationSignal);
    }
```
```java
    private final ThreadLocal<SQLiteSession> mThreadSession = new ThreadLocal<SQLiteSession>() {
        @Override
        protected SQLiteSession initialValue() {
            return createSession();
        }
    };
```

The problem is that SQLiteSession takes a connection from SQLiteConnectionPool for the duration of the transaction and keeps it in ThreadLocal variable.

So lets assume we are doing something in Thread1 and start a transaction. We take a lock in DefaultStoreIOSQLite with "synchronized (lock)" and proceed - the connection is taken from the pool and no connections are left there. The "synchronized (lock)"  is released.

Next a Thread2 comes and takes a lock "synchronized (lock)" and tries to take a connection by using "sqLiteOpenHelper.getWritableDatabase().beginTransaction();". Since there are no connections is starts to wait using "acquireConnection".

Then the Thread1 comes back with the work finished and tries to do "setTransactionSuccessful". "synchronized (lock)" is taken so he cannot do anything. "endTransaction" is never called and  the connection that was taken from the pool is never available again. Thread2 is locked and Thread1 is locked.

Deadlock.

IMO, it would be the best to remove your lock "synchronized (lock)" as the underlying SQLiteDatabase takes care of proper connection sharing by itself.

Thanks

artem-zinnatullin:
Good investigation, I tried to make deadlock with different implementations of `DefaultStorIOSQLite` before release 1.0.0 and current impl was LGTM, but your explanation makes sense.

I'll take a look at this tonight and provide solution, probably it will be possible to make it non blocking.

Thanks again!

artem-zinnatullin:
[...]

PR #482, opened 2015-08-05
Fix possible deadlock caused by internal synchronization in SQLiteDatabase

artem-zinnatullin:
Relolves #481.

At the moment, `StorIOSQLite` Operations already written in such way that they send notifications after transaction, so I just moved calls to `SQLiteDatabase` transactions out of the synchronization and keep code that manages notifications as is. 

@nikitin-da and @tadas-subonis please take a look and comment!

#deadlocksgoaway


tadas-subonis:
Well, it at least should solve the issue that I've reported earlier. Not sure about the notifications you mentioned earlier.

Thanks!

tadas-subonis:
It might be worth adding some tests using https://code.google.com/p/thread-weaver/wiki/UsersGuide

artem-zinnatullin:
We can not cover this with tests properly, because implementation of `SQLiteDatabase` may be changed in any version of Android SDK and we are using Robolectric for test with its own impl…

But, I'll think about more multithreading tests, our integration tests use multithreading BTW.

PR #491, opened 2015-08-16
Remove unnecessary synchronization, prevent possible deadlocks, faster & better

artem-zinnatullin:
Problem: we send notification to the `changesBus` inside of `synchronized(lock)` block and this notification may invoke user code that may open a db transaction and then wait for the `synchronized(lock)` and because of this db transaction, lock may never be released.

This PR fixes possible deadlock via moving notifications out of the `synchronized(lock)` block and also makes internal implementation faster.

@nikitin-da PTAL!

### DIFF
--- a/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/DefaultStorIOSQLite.java
+++ b/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/DefaultStorIOSQLite.java
@@ -20,10 +20,10 @@ import com.pushtorefresh.storio.sqlite.queries.UpdateQuery;
 import java.io.IOException;
 import java.util.HashMap;
 import java.util.HashSet;
-import java.util.Iterator;
 import java.util.Map;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import rx.Observable;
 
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
Example of deadlock:

1. `Thread1` makes `internal().notifyChange()` and this change throwed to the `changesBus` inside of `synchronized` block.
2. `Thread2` listens for changes and when change arrives he makes `internal.beginTransaction()` and aquires connection to the `SQLiteDabase` which prevents other Threads from accessing the db but then he want to increment `numberOfRunningTransactions` but Thread 1 holds the lock and does not allow `Thread2` to do this.

Problem: `Thread1` holds lock and `Thread2` is blocked on it and also it blocks the db, so no more changes to the db can be done.

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/DefaultStorIOSQLite.java
```diff
@@ -192,16 +192,14 @@ public DefaultStorIOSQLite build() {
...
-        private int numberOfRunningTransactions = 0;
+        @NonNull
+        private AtomicInteger numberOfRunningTransactions = new AtomicInteger(0);
 
         /**
          * Guarded by {@link #lock}.
          */
         @NonNull
-        private final Set<Changes> pendingChanges = new HashSet<Changes>();
+        private Set<Changes> pendingChanges = new HashSet<Changes>(5);
```
nikitin-da: 5?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/DefaultStorIOSQLite.java
```diff
@@ -378,22 +376,33 @@ public int delete(@NonNull DeleteQuery deleteQuery) {
...
-         */
         private void notifyAboutPendingChangesIfNotInTransaction() {
-            if (numberOfRunningTransactions == 0) {
-                final Iterator<Changes> iterator = pendingChanges.iterator();
+            final Set<Changes> changesToSend;
+
+            if (numberOfRunningTransactions.get() == 0) {
+                synchronized (lock) {
+                    changesToSend = pendingChanges;
+                    pendingChanges = new HashSet<Changes>(5);
```
nikitin-da: 5?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/DefaultStorIOSQLite.java
artem-zinnatullin: It's just randomly selected number used to avoid unnecessary memory allocation, I guess `5` is not big and not small for this, so, why not :)

nikitin-da:
I think, thread-safety hasn't broken. Great fix!

PR #501, opened 2015-08-30
- Person-Car-example: some discussion is needed

Rainer-Lang:


Rainer-Lang:
This is an example for an object with a list of other objects inside. 1-to-many

Rainer-Lang:
@artem-zinnatullin I want to load a person with all cars related - also update and delete. I also want to load, update and delete a car - with a given person.id

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonDeleteResolver.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class PersonDeleteResolver extends DeleteResolver<Person> {
+    @NonNull
+    @Override
+    public DeleteResult performDelete(@NonNull StorIOSQLite storIOSQLite, @NonNull Person person) {
+        // We can even reuse StorIO methods
+        storIOSQLite
+                .delete()
+
+                // TODO here I have a problem with Person-class: Is this right?
```
artem-zinnatullin: yep, you can do this :)

artem-zinnatullin: Oh, I got it, it's recursive…

Rainer-Lang:
@artem-zinnatullin This is only a small and simple example. I use complexer classes with more relations inside (up to 4 lists of other tables/classes) - but if the simple example work - all other will also work - I hope. ;-)

artem-zinnatullin:
@Rainer-Lang I am very sorry, way tooooo busy this week, I'll look as soon I get some free time, sorry :(

Rainer-Lang:
@artem-zinnatullin Thanks for the informations. I'm looking forward to discuss the example and see how this could be done with StorIO.

Rainer-Lang:
@artem-zinnatullin fyi: I also use the same objects GSON parsing from a webservice.

artem-zinnatullin:
@Rainer-Lang I'm here, reviewing

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Car.java
```diff
@@ -0,0 +1,66 @@
...
+    /**
+     * If object was not inserted into db, id will be null
+     */
+    @Nullable
+    @StorIOSQLiteColumn(name = CarsTable.COLUMN_ID, key = true)
+    Long id;
+
+    @Nullable
+    @StorIOSQLiteColumn(name = CarsTable.COLUMN_ID_PERSON)
+    Long idPerson;
```
artem-zinnatullin: Why `car` has an Id of a person?

Rainer-Lang:
The car has the id of the person because a car ja an owner.  A person could have many cars but a car could only have one owner. 
It's a 1 to many relation.

artem-zinnatullin:
Okay, my bad.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
```diff
@@ -0,0 +1,68 @@
...
+
+import com.pushtorefresh.storio.sample.db.tables.PersonsTable;
+import com.pushtorefresh.storio.sqlite.annotations.StorIOSQLiteColumn;
+import com.pushtorefresh.storio.sqlite.annotations.StorIOSQLiteType;
+
+import java.util.ArrayList;
+import java.util.List;
+
+@StorIOSQLiteType(table = PersonsTable.TABLE)
+public class Person {
```
artem-zinnatullin: Since this class has custom operation resolvers you don't need annotations

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonDeleteResolver.java
artem-zinnatullin: 1. Open a transaction via `storIOSQLite.internal.beginTransaction()`
2. Delete a `person` via `storIOSQLite.internal().delete()`
3. Delete `cars` via `storIOSQLite.delete().objects(cars)`
4. Close the transaction

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java
```diff
@@ -0,0 +1,39 @@
...
+    @NonNull
+    @Override
+    public Person mapFromCursor(@NonNull Cursor cursor) {
+        final Person person = Person.newPerson(
+                cursor.getLong(cursor.getColumnIndexOrThrow(UsersTable.COLUMN_ID)),
+                cursor.getString(cursor.getColumnIndexOrThrow(UsersTable.COLUMN_NICK))
+        );
+
+        List<Car> cars = new ArrayList<Car>();
+        while (cursor.moveToNext()) {
```
artem-zinnatullin: I think you need to do separate request to the db to retrieve all `cars` of this person

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonPutResolver.java
```diff
@@ -0,0 +1,43 @@
...
+
+    @NonNull
+    @Override
+    public PutResult performPut(@NonNull StorIOSQLite storIOSQLite, @NonNull Person person) {
+        // We can even reuse StorIO methods
+        final PutResults<Object> putResults = storIOSQLite
+                .put()
+
+                // TODO here I need help
+                .objects(asList(person, person.getCars()))
```
artem-zinnatullin: You can not put `person` via `storIOSQLite.put()` because it will be recursive, so you need to use `storIOSQLite.internal()`, but you can use `storIOSQLite.put()` to put `cars`.

Rainer-Lang:
@artem-zinnatullin Thanks. Now I look what I can do and after changes I will post the example again. Maybe then it's working... I hope - we will see. :-)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: Really none(!) annotations? Also not for the primary key and table?

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonPutResolver.java
Rainer-Lang: ```storIOSQLite.internal()``` and then use ```insert``` or ```update```. Right? But then I also have to use an ```uri``` and I haven't an ```uri```.....

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
artem-zinnatullin: Annotations will only trigger generation of Put, Get and Delete Operation resolvers in compile time, since you don't use autogenerated code for `Person` class — you don't need annotations :)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonPutResolver.java
artem-zinnatullin: Yep, why would you need `uri` for `StorIOSQLite`? I guess you also added `storio:content-resolver` to the dependencies :)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonDeleteResolver.java
```diff
@@ -0,0 +1,50 @@
...
+
+import java.util.HashSet;
+import java.util.Set;
+
+import static java.util.Arrays.asList;
+
+public class PersonDeleteResolver extends DeleteResolver<Person> {
+    @NonNull
+    @Override
+    public DeleteResult performDelete(@NonNull StorIOSQLite storIOSQLite, @NonNull Person person) {
```
artem-zinnatullin: Now it's correct! :+1:

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonPutResolver.java
Rainer-Lang: Yes - you're completely right. I imported a ```content-resolver``` method.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java
Rainer-Lang: Could you please give some more informations how to do this?

Rainer-Lang:
@artem-zinnatullin Should I make another pr? I think this one is a bit messed up...

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java
artem-zinnatullin: Sure :)

As I see: one `person` may have multiple `cars` so in general you have 2 ways of retrieving Person + cars from the db:

1. Make query with `join` which will retrieve rows (`person`, `car`), but the problem is that if one `person` has multiple `cars` you will have duplicates of this `person` in next `n` rows, like this: `row1` ==  `person1`, `car1`, `row2` == `person1`, `car2`, `row3` == `person2`, `car3` and so on. It's possible and fast (for the db) way to retrieve `persons` and their `cars`.
2. Make query to only `persons` table and then attach `cars` for each `person` programmatically, but it will require a lot of queries to the `cars` table, which makes this solution slower than 1rst one

Now you need to decide which way you want to go :)

artem-zinnatullin:
>Should I make another pr? I think this one is a bit messed up...

No, let's work in this one, we can squash commits later, it's not a problem :)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java
Rainer-Lang: Solution 2 is an outer join, right?

But wait: I only have to use the ```person.id``` in the ```cars``` table. That should be a simple query, right? :-)

Rainer-Lang:
```final List<Car> cars = storIOSQLite
  .get()
  .listOfObjects(Car.class)
  .withQuery(Query.builder()
    .table("cars")
    .build())
  .prepare()
  .executeAsBlocking();```

Comment (artem-zinnatullin) [inline storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java]: > Solution 2 is an outer join, right?

Nope, Solution 2 is a separate query to the `cars` table for each `person`, your code is correct :)

To get `storIOSQLite` instance you need to override `GetResolver` instead of `DefaultGetResolver` (Default resolvers were made for simplicity, but you always can override most abstract resolvers to write any kind of custom code :)

Comment (Rainer-Lang) [inline storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java]: It's a bit tricky I see.... Cursor here Object (Person) there, Query on other place... :-)

Comment (artem-zinnatullin) [inline storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java]: Yep, but with flexible design of StorIO everything is possible!

Comment (Rainer-Lang) [inline storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/PersonGetResolver.java]:
```java
public class PersonGetResolver extends GetResolver<Person> {

    @NonNull
    @Override
    public Cursor performGet(@NonNull StorIOSQLite storIOSQLite, @NonNull RawQuery rawQuery) {

        // TODO should be used to get persons AND cars

        RawQuery rQ = RawQuery.builder()
                .query("select p._id, p.name, c._id, c.model from persons p left join cars c on p.id=c.id_person")
                .build();

        return storIOSQLite.internal().rawQuery(rQ);
    }


    @NonNull
    @Override
    public Cursor performGet(@NonNull StorIOSQLite storIOSQLite, @NonNull Query query) {

        // TODO should not be used to get persons AND cars

        return storIOSQLite.internal().query(query);
    }

    @NonNull
    @Override
    public Person mapFromCursor(@NonNull Cursor cursor) {
        final Person person = Person.newPerson(
                cursor.getLong(cursor.getColumnIndexOrThrow(UsersTable.COLUMN_ID)),
                cursor.getString(cursor.getColumnIndexOrThrow(UsersTable.COLUMN_NICK))
        );

        List<Car> cars = new ArrayList<>();
        while (cursor.moveToNext()) {
            cars.add(Car.newCar(
                    cursor.getLong(cursor.getColumnIndexOrThrow(CarsTable.COLUMN_ID)),
                    cursor.getString(cursor.getColumnIndexOrThrow(CarsTable.COLUMN_MODEL)),
                    person.getId()
            ));
        }

        person.setCars(cars);

        return person;
    }
}
```

Rainer-Lang: I love the sentence: "with flexible design of StorIO everything is possible"
:-)

artem-zinnatullin: With posted code you will only get one person with all his cars, or you don't need to retrieve all persons via one request?

Rainer-Lang: And then if I want to get a ```Person``` I have to use this?

```java
final Subscription subscription = storIOSQLite
                .get()
                .listOfObjects(Person.class)
                .withQuery(RawQuery.builder()
                  .query("select p._id, p.name, c._id, c.model from persons p left join cars c on p.id=c.id_person")
                  .build()
                .prepare()
                .createObservable()
...
```

Rainer-Lang: I want to get (a) one person for an id and also (b) ALL persons with ALL cars.

artem-zinnatullin: Okay, so I'd suggest you to write `PersonGetResolver` in such manner as if you would work with it just for querying `person` table, so you would pass each query directly to the db and then just iterate over each `person` and retrieve his `cars`.

So you will be able to write any possible queries to the `persons` table and it'll work!

Try to change code to this and then I'll suggest ways of improving it :)

Rainer-Lang:
Now new commit:
- Person.getCars() rewritten to load cars from db
- PersonGetResolver - inherited from DefaultGetResolver to handle Person.get with all queries
- PersonCarsFragment.reloadData rewritten to iterate over all loaded persons and load all cars

Please optimize eveything you will find! :-)

Rainer-Lang:
Please also have a look at ```PersonPutResolver``` at ```performPut()``` and ```putResults```.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
```diff
@@ -0,0 +1,97 @@
...
+
+    public void setName(@NonNull String name) {
+        this.name = name;
+    }
+
+//    @NonNull
+//    public List<Car> getCars() { return cars; }
+
+    @NonNull
+    public List<Car> getCars(StorIOSQLite storIOSQLite) {
```
artem-zinnatullin: I am not sure that this is good idea, you probably don't want to do blocking I/O when  you read the property of the object.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
```diff
@@ -0,0 +1,97 @@
...
+        if ( storIOSQLite != null && cars == null ) {
+            cars = new ArrayList<>();
+
+            final Subscription subscription = storIOSQLite
```
artem-zinnatullin: Why you decided to go with Rx here? `executeAsBlocking()` will be faster!

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
```diff
@@ -0,0 +1,97 @@
...
+                            .build())
+                    .prepare()
+                    .createObservable() // Get Result as rx.Observable and subscribe to further updates of tables from Query!
+                    .observeOn(AndroidSchedulers.mainThread()) // All Rx operations work on Schedulers.io()
+                    .subscribe(cars -> this.cars = cars, throwable -> Timber.e(throwable, "getCars()")
+                    );
+            subscription.unsubscribe();
+        }
+
+        return cars;
```
artem-zinnatullin: Actually, here you're returning empty list which will be changed in some `Schedulers.io()` thread later, this code is absolutely non-thread safe!

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Queen.java
```diff
@@ -0,0 +1,163 @@
...
+
+import rx.Observer;
+import rx.Subscription;
+import rx.android.schedulers.AndroidSchedulers;
+
+import static com.pushtorefresh.storio.sample.ui.Toasts.safeShowShortToast;
+import static rx.android.schedulers.AndroidSchedulers.mainThread;
+
+@StorIOSQLiteType(table = QueensTable.TABLE)
+public class Queen {
```
artem-zinnatullin: I thought we were discussing Persons and Cars :)

artem-zinnatullin:
@Rainer-Lang if you don't mind, I can just rewrite your example with `Persons` and `Cars` from scratch as max optimized as possible, and then you'll ask me things that you would like to ask, what do you think about it? :)

Rainer-Lang:
I think I could learn very much from your code and I would like to see how StorIO is used in the right way. 
So my answer is yes :-)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: But ```executeAsBlocking``` is blocking the db for other inserts or updates, right? Just to know.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: You're right, I don't want to block ui-thread.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: Could give suggestions please?

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Queen.java
Rainer-Lang: Yes of course! :-D
That's only another example - made only for me. Don't look at it. ;-)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
artem-zinnatullin: You need to make blocking request to the db, fill the list and only then return it.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
artem-zinnatullin: `blocking` means that it will be executed synchronously on the current Thread, how it will be dispatched by the SQLite is not related to StorIO :)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: So I have to use ```executeAsBlocking``` here?
I have to get the cars of a person in the same thread, right? Therefore "blocking".

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: Got it. Thanks Artem. :-)
Looking forward to your full rewrite of this person-cars-example.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
artem-zinnatullin: Yep, or make blocking call via Rx if you want to use it here (I don't see need in this, but it's possible):

```java
storIOSQLite
  .get()
  ...
  .createObservable()
  .toBlocking()
  .first()
```

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
artem-zinnatullin: I'll try to do it this week (a lot of work + I have to visit hospital every day this week…)

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: OK.
If I want to execute ```getCars()``` later then I think it's better to have an Observable.

storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/entities/Person.java
Rainer-Lang: Wrote an email.

artem-zinnatullin:
Closing, feel free to comment :)

PR #512, opened 2015-09-10
Remove coloumn and table names from parameters

nikitin-da:
Database identifiers don't allow to use as SQL parameters/placeholders

@artem-zinnatullin PTAL


artem-zinnatullin:
LGTM

PR #514, opened 2015-09-10
adding somebytes column that was missing from CREATE TABLE TweetsTable

cpeppas:
This is needed for the sample app to display tweets


artem-zinnatullin:
@cpeppas thanks! It's my fail, was experimenting with the different types of columns and committed wrong code.

I guess, we should add integration tests to the sample app to assert that everything work fine!

Thanks! (I'll remove this column in future since it's not required for `Tweet`, but for now let's merge this)

PR #524, opened 2015-09-18
Remove nulling ThreadLocal since it's already WeakReference

artem-zinnatullin:
It could lead to `NullPointerException` in some cases.

@nikitin-da PTAL :)


nikitin-da:
:+1:

PR #531, opened 2015-09-23
Add @CheckResult annotation, makes life in the Android Studio Better!

artem-zinnatullin:
@nikitin-da PTAL :)


nikitin-da:
It seems flakiness decided to make it last time

PR #543, opened 2015-10-15
Add API for observing all changes in StorIOSQLite

### LINKED ISSUE #446: Provide public API to ChangesBus to allow users subscribe to all changes in StorIO
At the moment user can observe changes of particular tables (in `StorIOSQLite`) or particular `Uris` (in `StorIOContentResolver`).

I think it could be useful for users (at least for me and our current work project) to have an API that will allow to observe all changes without filtration, for example for logging and other purposes.

@nikitin-da thoughts? I remember that we discussed this several months ago and decided to provide only `observeChanges(tables)` and `observeChanges(uris)` to prevent users from EventBus-like code, but now I see usages for general `observeChanges()` API.

artem-zinnatullin:
Closes #446!

@nikitin-da PTAL :)


nikitin-da:
:+1:

PR #561, opened 2015-11-17
Android Gradle Plugin 1.5.0

artem-zinnatullin:
PTAL


artem-zinnatullin:
@nikitin-da ptal :)

nikitin-da:
:+1:

PR #586, opened 2016-01-13
Remove "final" from most of the classes

artem-zinnatullin:
This will allow mocking StorIO because at the moment we have to cheat and use real db via Robolectric in tests.

// Also, I'm slowly working on `MockStorIOSQLite`

@nikitin-da PTAL :)


nikitin-da:
Ook

PR #593, opened 2016-01-17
Enable emails from Travis to react on problems with master branch

artem-zinnatullin:
Previously we have status checks from GitHub which forces us to update each branch with master after each new changes in master which was a problem with many PRs.

Now I'll receive an email if PR merged into master will break the build and PRs now can be based on non-latest changes in master.

@nikitin-da PTAL

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,12 @@ script:
   - sh ci.sh
 
 notifications:
-  email: false
+  email:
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
@nikitin-da I've added your email too :)

nikitin-da:
Thank you)

PR #596, opened 2016-01-18
Test asRxObservable() instead of createObservable() which is now deprecated

artem-zinnatullin:
@nikitin-da PTAL :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ storIOSQLite
     .table("tweets")
     .build())
   .prepare()
-  .createObservable() // Get Result as rx.Observable and subscribe to further updates of tables from Query!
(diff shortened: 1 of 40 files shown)


README.md
```diff
@@ -104,7 +104,7 @@ storIOSQLite
     .table("tweets")
     .build())
   .prepare()
-  .createObservable() // Get Result as rx.Observable and subscribe to further updates of tables from Query!
+  .asRxObservable() // Get Result as rx.Observable and subscribe to further updates of tables from Query!
```
artem-zinnatullin: Lol, here we literaly saying `Get Result as rx.Observable` :smile: 

So, I guess, new method name is okay :)

nikitin-da: Hehe

PR #608, opened 2016-01-27
Deprecate StorIOContentResolver.Internal and add alias LowLevel

### LINKED ISSUE #516: Rename Internal to LowLevel
`Internal` is a name for something that should not be used by others, but we allow users to use our `StorIOSQLite.Internal` and `StorIOContentResolver.Internal` APIs for Operation resolvers and so on, so I guess, we should rename it to make things clearer.

@nikitin-da and anybody who interested in this, what is your opinion?

artem-zinnatullin:
Closes #516.

@nikitin-da PTAL carefully :)


nikitin-da:
Looks good, but it we still use Internal in storIOSQLite tests =(

PR #610, opened 2016-01-29
Fix Relations in sample app (JOIN column names collision)

### LINKED ISSUE #603: Incorrect mapping from JOINs when there are columns with the same name
Using [this get resolver](https://github.com/pushtorefresh/storio/blob/master/storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/db/resolvers/TweetWithUserGetResolver.java) with [this query](https://github.com/pushtorefresh/storio/blob/master/storio-sample-app/src/main/java/com/pushtorefresh/storio/sample/sample_code/Relations.java) would map the tweet id as the user id because `TweetsTable.COLUMN_ID` and `UsersTable.COLUMN_ID` are the same (`_id`).

One solution would be to select one of them with another value (like `SELECT tweets._id AS tweetId, *`):

    public TweetWithUser mapFromCursor(@NonNull Cursor cursor) {
        final Tweet tweet = Tweet.newTweet(
                cursor.getLong(cursor.getColumnIndexOrThrow("tweetId")),
                cursor.getString(cursor.getColumnIndexOrThrow(TweetsTable.COLUMN_AUTHOR)),
                cursor.getString(cursor.getColumnIndexOrThrow(TweetsTable.COLUMN_CONTENT))
        );

        final User user = User.newUser(
                cursor.getLong(cursor.getColumnIndexOrThrow(UsersTable.COLUMN_ID)),
                cursor.getString(cursor.getColumnIndexOrThrow(UsersTable.COLUMN_NICK))
        );

        return new TweetWithUser(tweet, user);
    }

Another one is using the foreign key that references the other table after the mapping (`tweet.id = user.tweet_id`). I've used this one to reutilize `mapFromCursor` for both entities, but the order from `JOIN` matters:

    public TweetWithUser mapFromCursor(@NonNull Cursor cursor) {
        final Tweet tweet = tweetGetResolver.mapFromCursor(cursor);
        final User user = userGetResolver.mapFromCursor(cursor);
        tweet.id = user.tweet_id;

        return new TweetWithUser(tweet, user);
    }

I don't think StorIO is able to workaround this, but I think the sample code should warn about this because it can mess the database when updating it with incorrect fields.

Edit: The second approach is for a 1-n relationship with foreign keys, it does not apply to the actual `Tweet` and `User` examples.

artem-zinnatullin:
Fixes #603.

### DIFF
--- a/storio-sample-app/src/test/java/com/pushtorefresh/storio/sample/sample_code/RelationsTest.java
+++ b/storio-sample-app/src/test/java/com/pushtorefresh/storio/sample/sample_code/RelationsTest.java
@@ -1,0 +1,49 @@
+package com.pushtorefresh.storio.sample.sample_code;
+
+import com.pushtorefresh.storio.contentresolver.BuildConfig;
+import com.pushtorefresh.storio.sample.SampleApp;
+import com.pushtorefresh.storio.sample.db.entities.Tweet;
+import com.pushtorefresh.storio.sample.db.entities.TweetWithUser;
+import com.pushtorefresh.storio.sample.db.entities.User;
+
+import org.junit.Test;
+import org.junit.runner.RunWith;
+import org.robolectric.RobolectricGradleTestRunner;
+import org.robolectric.RuntimeEnvironment;
+import org.robolectric.annotation.Config;
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(RobolectricGradleTestRunner.class)
+@Config(constants = BuildConfig.class, sdk = 21)
+public class RelationsTest {
+
+    @Test
+    public void name() {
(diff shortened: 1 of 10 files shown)


storio-sample-app/src/test/java/com/pushtorefresh/storio/sample/sample_code/RelationsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import org.robolectric.annotation.Config;
+
+import java.util.List;
+
+import static java.util.Arrays.asList;
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(RobolectricGradleTestRunner.class)
+@Config(constants = BuildConfig.class, sdk = 21)
+public class RelationsTest {
```
nikitin-da: :metal:

nikitin-da:
Looks good

PR #618, opened 2016-02-24
Fix syntax error in the ContentResolverTypeMapping builder

roughike:
The <Tweet> type mapping is on the wrong side of the builder() method, and caused some confusion / frustration to a friend of a friend of a friend of mine. But of course not me. I'm better than that.


artem-zinnatullin:
Oh damn, my bad, this syntax is mess in Java, thanks! 

// Awesome description lol

PR #626, opened 2016-03-07
Add tests to check inserting and updating with nullable fields

### LINKED ISSUE #624: Updating a row with null value
I don't know if I am missing something but if I try to update a row with some columns as null, it doesn't seem to be reflected

nikitin-da:
Closes #624 
@artem-zinnatullin and @sreejithraman PTAL, please
All 3 variants are works via content resolver and sqlite directly:
- insert `null`
- update from `null` to `not null`
- update from `not null` to `null`


artem-zinnatullin:
:+1: awesome, thanks a lot!

PR #642, opened 2016-04-23
Add ignoreNull property

### LINKED ISSUE #595: Better support for fields nullability by Annotation Processors
public ContentValues mapToContentValues(@NonNull ReplyBean object) {
        ContentValues contentValues = new ContentValues(2);

        if(object.content != null){
            contentValues.put("content", object.content);
        }
        if(object.key!= null){
            contentValues.put("key", object.key);
        }
       
        return contentValues;
    }

First I appreciate your work, I think It will be helpful to check the properties whether they are null, and the properties should be non-primitive

nikitin-da:
Closes #595

### DIFF
--- a/storio-content-resolver-annotations-processor/src/main/java/com/pushtorefresh/storio/contentresolver/annotations/processor/generate/PutResolverGenerator.java
+++ b/storio-content-resolver-annotations-processor/src/main/java/com/pushtorefresh/storio/contentresolver/annotations/processor/generate/PutResolverGenerator.java
@@ -19,7 +19,7 @@ import static com.pushtorefresh.storio.common.annotations.processor.generate.Com
 import static javax.lang.model.element.Modifier.PROTECTED;
 import static javax.lang.model.element.Modifier.PUBLIC;
 
-public class PutResolverGenerator  implements Generator<StorIOContentResolverTypeMeta> {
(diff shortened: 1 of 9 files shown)


storio-content-resolver-annotations-processor/src/main/java/com/pushtorefresh/storio/contentresolver/annotations/processor/generate/PutResolverGenerator.java
```diff
@@ -115,11 +115,20 @@ private MethodSpec createMapToContentValuesMethodSpec(@NotNull final StorIOConte
                 .addCode("\n");
 
         for (final StorIOContentResolverColumnMeta columnMeta : storIOContentResolverTypeMeta.columns.values()) {
+            final boolean primitive = columnMeta.element.asType().getKind().isPrimitive();
+            final boolean needIgnoreNull = columnMeta.storIOColumn.ignoreNull();
```
rsinukov: does it make sense to throw or at least log if user annotated primitive type with @IgnoreNull?

nikitin-da: Oh, I agree, I'll throw exception here

nikitin-da:
2 oks from #643

PR #629, opened 2016-03-14
rx.Completable support for write operations

geralt-encore:
https://github.com/pushtorefresh/storio/issues/628

PR #654, opened 2016-05-18
Release 1.9.0

nikitin-da:


artem-zinnatullin:
Oh nice! I remember that we had some issue/feature that was half-done and committed but not sure about it, don't you see anything like this? 

Just afraid that we can ship half-completed state of feature/issue.

nikitin-da:
Hm, I don't see such PR's merged in master. Just #629 looks like by name, but put and delete operation is whole list of completable operations.

artem-zinnatullin:
ok, 👍 from me, let's wait for review from @karlicoss and I'll publish it on maven central!

Btw, I'm totally ok with frequent releases, like each 1/2/3/4/whatever weeks, should be good for users and library.

// Many thanks to you guys, I'm sorry that I can't actively participate in library development at the moment.

artem-zinnatullin:
@karlicoss any input?

karlicoss:
@artem-zinnatullin @nikitin-da looks good, let's ship it!

PR #656, opened 2016-05-19
Update RxJava to 1.1.5

artem-zinnatullin:
Changelog https://github.com/ReactiveX/RxJava/blob/1.x/CHANGES.md#version-115---may-5-2016-maven

Just want to be sure that StorIO works with RxJava 1.1.5 correctly.


nikitin-da:
Works 👍

PR #657, opened 2016-05-19
Update Dagger for sample app to 2.4

artem-zinnatullin:
Changelog https://github.com/google/dagger/blob/master/CHANGELOG.md#version-24-2016-04-21


nikitin-da:
Ok

PR #665, opened 2016-06-29
Update gradle version to 2.1.2

geralt-encore:
Updated gradle version, because 1.5 is quite outdated.
Sdk manager plugin doesn't work with gradle 2.0+ ([source](https://github.com/JakeWharton/sdk-manager-plugin/issues/99)). Maybe there is a better solution?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -10,9 +10,6 @@ buildscript {
         // Android Gradle plugin
         classpath 'com.android.tools.build:gradle:2.1.2'
 
-        // Resolves android-sdk dependencies
-        classpath 'com.github.JakeWharton:sdk-manager-plugin:master'
-
         // APT compile-time annotation processing
         classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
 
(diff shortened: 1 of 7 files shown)


artem-zinnatullin:
Let's remove SDK manager plugin, it doesn't really work and deprecated by Jake.

build.gradle
```diff
@@ -3,14 +3,12 @@
 buildscript {
     repositories {
         jcenter()
+        maven { url 'https://jitpack.io' }
```
nikitin-da: It seems it is not required ;)

nikitin-da:
👍

artem-zinnatullin:
👍

nikitin-da:
Thanks, @geralt-encore!

oldergod:
@artem-zinnatullin How will you deal with CI now that you removed the SDK manager ?

artem-zinnatullin:
Just declare all dependencies in .travis.yml/circle.yml

Issue #381, opened 2015-06-13
Add Infer to the CI script


artem-zinnatullin:
https://code.facebook.com/posts/1648953042007882/open-sourcing-facebook-infer-identify-bugs-before-you-ship/

PR #666, opened 2016-06-29
Add Infer plugin

geralt-encore:
https://github.com/pushtorefresh/storio/issues/381

I've added [this](https://github.com/uber-common/infer-plugin) Gradle plugin, which plays with Infer nicely.
It runs ok on my local machine. But we still need to install Infer to the CI server (not sure how) and then add two Gradle tasks to the CI script: `inferDebug` and `eradicateDebug` (for [Eradicate](http://fbinfer.com/docs/eradicate.html) feature).
Right now Infer doesn't see any issues (for testing purposes I've added dummy method with a silly null error and it has found it - so it actually works). Eradicate is finding one issue in storio-common module and fails right after that, so I can't say anything about other modules (can just run eradicate on separate modules, but haven't done it yet). If you are interested in applying these checks I could fix them before the merge.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -125,6 +125,9 @@ subprojects {
                     exceptionFormat 'full'
                 }
             }
+
+            // apply infer plugin
+            project.apply plugin: 'com.uber.infer.android'
         }
     }
 }
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ buildscript {
         // https://github.com/palantir/gradle-jacoco-coverage
         classpath 'com.palantir:jacoco-coverage:0.2.0'
 
-        classpath "com.uber:infer-plugin:0.7.1"
+        classpath 'com.uber:infer-plugin:0.7.1'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
--- a/build.gradle
+++ b/build.gradle
@@ -17,6 +17,8 @@ buildscript {
         // https://github.com/palantir/gradle-jacoco-coverage
         classpath 'com.palantir:jacoco-coverage:0.2.0'
 
+        classpath "com.uber:infer-plugin:0.7.1"
+
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
     }
--- a/storio-common/build.gradle
+++ b/storio-common/build.gradle
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.library'
 apply plugin: 'checkstyle'
+apply plugin: 'com.uber.infer.android'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
--- a/storio-common/build.gradle
+++ b/storio-common/build.gradle
@@ -1,6 +1,5 @@
 apply plugin: 'com.android.library'
 apply plugin: 'checkstyle'
-apply plugin: 'com.uber.infer.android'
 
 android {
     compileSdkVersion rootProject.ext.compileSdkVersion
--- a/storio-content-resolver-annotations-processor/build.gradle
+++ b/storio-content-resolver-annotations-processor/build.gradle
@@ -1,4 +1,5 @@
 apply plugin: 'java'
+apply plugin: 'com.uber.infer.java'
 
 targetCompatibility = '1.6'
 sourceCompatibility = '1.6'
--- a/storio-content-resolver-annotations-processor/build.gradle
+++ b/storio-content-resolver-annotations-processor/build.gradle
@@ -1,5 +1,4 @@
 apply plugin: 'java'
-apply plugin: 'com.uber.infer.java'
 
 targetCompatibility = '1.6'
 sourceCompatibility = '1.6'
--- a/storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/delete/PreparedDeleteCollectionOfObjects.java
+++ b/storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/delete/PreparedDeleteCollectionOfObjects.java
@@ -56,7 +56,7 @@ public class PreparedDeleteCollectionOfObjects<T> extends PreparedDelete<DeleteR
      *
      * @return non-null results of Delete Operation.
      */
-    @SuppressWarnings("unchecked")
+    @SuppressWarnings({"unchecked", "ConstantConditions"})
     @WorkerThread
     @NonNull
     @Override
@@ -96,7 +96,7 @@ public class PreparedDeleteCollectionOfObjects<T> extends PreparedDelete<DeleteR
                     final DeleteResult deleteResult = explicitDeleteResolver.performDelete(storIOContentResolver, object);
                     results.put(object, deleteResult);
                 }
-            } else {
+            } else if (objectsAndDeleteResolvers != null) {
                 for (final SimpleImmutableEntry<T, DeleteResolver<T>> objectAndDeleteResolver : objectsAndDeleteResolvers) {
                     final T object = objectAndDeleteResolver.getKey();
                     final DeleteResolver<T> deleteResolver = objectAndDeleteResolver.getValue();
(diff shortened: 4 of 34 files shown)


artem-zinnatullin:
But you don't run it during CI build. See https://github.com/artem-zinnatullin/qualitymatters/issues/21 :(

geralt-encore:
That sucks =(
I've found only [this](https://github.com/facebook/infer/issues/153) solution, but it looks really bad.
Maybe it is an option to you it just manually? Better than nothing at least and could be helpful sometimes.

artem-zinnatullin:
Okay, I'm fine with running it locally

geralt-encore:
All eradicate warnings are fixed now

nikitin-da:
Nice! 
👍 but covarage

geralt-encore:
Yeah, but there is not much that I can do since I've added only annotation and couple of null checks

nikitin-da:
Hm, I think false branches of that checks weren't covered =) https://codecov.io/gh/pushtorefresh/storio/compare/29a3b1348fd8e0b8d8b49227dd4288454e509099...7529dfe424c59a30bb21de67886bc05f9e484d92

geralt-encore:
But most of these ifs are redundant actually and mostly about making infer stop complaining about them. So there is not much sense for covering them.

build.gradle
```diff
@@ -17,6 +17,8 @@ buildscript {
         // https://github.com/palantir/gradle-jacoco-coverage
         classpath 'com.palantir:jacoco-coverage:0.2.0'
 
+        classpath "com.uber:infer-plugin:0.7.1"
```
artem-zinnatullin: single quotes pls

storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/delete/PreparedDeleteCollectionOfObjects.java
```diff
@@ -96,7 +96,7 @@
                     final DeleteResult deleteResult = explicitDeleteResolver.performDelete(storIOContentResolver, object);
                     results.put(object, deleteResult);
                 }
-            } else {
+            } else if (objectsAndDeleteResolvers != null) {
```
artem-zinnatullin: no, it's ok, we better crash if it'll be null

artem-zinnatullin: just suppress it here

storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/put/PutResults.java
```diff
@@ -65,7 +65,7 @@ public int numberOfInserts() {
         int numberOfInserts = 0;
 
         for (T object : results.keySet()) {
-            if (results.get(object).wasInserted()) {
+            if (results.get(object) != null && results.get(object).wasInserted()) {
```
artem-zinnatullin: result can never be null, we don't need this `!= null` here

storio-content-resolver/src/main/java/com/pushtorefresh/storio/contentresolver/operations/put/PutResults.java
```diff
@@ -92,7 +92,7 @@ public int numberOfUpdates() {
         for (T object : results.keySet()) {
             final PutResult putResult = results.get(object);
 
-            if (putResult.wasUpdated()) {
+            if (putResult != null && putResult.wasUpdated() && putResult.numberOfRowsUpdated() != null) {
```
artem-zinnatullin: result can never be null, we don't need this `!= null` here

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/delete/PreparedDeleteCollectionOfObjects.java
```diff
@@ -115,7 +115,7 @@
                             lowLevel.notifyAboutChanges(Changes.newInstance(deleteResult.affectedTables()));
                         }
                     }
-                } else {
+                } else if (objectsAndDeleteResolvers != null){
```
artem-zinnatullin: same as above

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/put/PreparedPutCollectionOfObjects.java
```diff
@@ -108,7 +108,7 @@
                             lowLevel.notifyAboutChanges(Changes.newInstance(putResult.affectedTables()));
                         }
                     }
-                } else {
+                } else if (objectsAndPutResolvers != null) {
```
artem-zinnatullin: same as above

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/put/PutResults.java
```diff
@@ -65,7 +65,7 @@ public int numberOfInserts() {
         int numberOfInserts = 0;
 
         for (T object : results.keySet()) {
-            if (results.get(object).wasInserted()) {
+            if (results.get(object) != null && results.get(object).wasInserted()) {
```
artem-zinnatullin: result can never be null, we don't need this `!= null` here

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/put/PutResults.java
```diff
@@ -92,7 +92,7 @@ public int numberOfUpdates() {
         for (T object : results.keySet()) {
             final PutResult putResult = results.get(object);
 
-            if (putResult.wasUpdated()) {
+            if (putResult != null && putResult.wasUpdated() && putResult.numberOfRowsUpdated() != null) {
```
artem-zinnatullin: result can never be null, we don't need this `!= null` here

geralt-encore:
@artem-zinnatullin the thing is that it is impossible to suppress warnings on infer side, so without these redundant checks build is failing. Looks like infer is not smart enough for some cases.

artem-zinnatullin:
@geralt-encore I definitely don't like that it changes logic in some places and instead of failing with NPE we're changing behavior.

What if you add plain old java `assert value != null`? It won't be executed on Android but will be executed in our JVM tests and probably will prevent Infer from complaining

geralt-encore:
Yeah, I completely agree with you. I'll try assert solution and if it won't work we will probably have to abandon this idea =(

geralt-encore:
Looks like assert stuff worked. I'll rework it then in a short time.

artem-zinnatullin:
ok, great!

storio-common/build.gradle
```diff
@@ -1,5 +1,6 @@
 apply plugin: 'com.android.library'
 apply plugin: 'checkstyle'
+apply plugin: 'com.uber.infer.android'
```
artem-zinnatullin: Maybe apply it in root `build.gradle` for all `android` subprojects?

See https://github.com/pushtorefresh/storio/blob/592368c97524f736e315d7e89cf4f2b339c657b7/build.gradle#L107

storio-content-resolver-annotations-processor/build.gradle
```diff
@@ -1,4 +1,5 @@
 apply plugin: 'java'
+apply plugin: 'com.uber.infer.java'
```
nikitin-da: What about common and sqlite annotation processor modules? ;)

geralt-encore: Thank you, I just forgot to remove it from here. It doesn't work with annotation processors at all.
```
* What went wrong:
Execution failed for task ':storio-content-resolver-annotations-processor:inferPrepare'.
> Configuration with name 'provided' not found.
```

geralt-encore:
I have no idea what is the problem with `PutResultsTest` `checkToStringImplementation`. Something relate to assertions I guess.

artem-zinnatullin:
lol wtf

artem-zinnatullin:
Got it, your pr is #666

geralt-encore:
Damn knew it would be a problem 😬

artem-zinnatullin:
Still fails :(

geralt-encore:
I have no idea how can I possibly fix it

geralt-encore:
@artem-zinnatullin probably we can close it since no one has an idea what's wrong with failing test and it's not really important one.

artem-zinnatullin:
Yeah, ok :(

PR #674, opened 2016-07-28
Make generated map methods public

### LINKED ISSUE #673: Make public generated mapToInsertQuery, mapToUpdateQuery,...  methods
For instance in the : https://github.com/pushtorefresh/storio/blob/master/storio-sqlite-annotations-processor/src/main/java/com/pushtorefresh/storio/sqlite/annotations/processor/generate/PutResolverGenerator.java
Modifiers are put as public for mapToContentValues, but not for mapToInsertQuery, mapToUpdateQuery, ...

It would be useful for when using storIOSQLite.lowLevel().insert() or .insertWithOnConflict() to be able to retrieve them.

I can do the PR for that if you want.

nikitin-da:
Closes #673


artem-zinnatullin:
Hm, maybe make them `final`…?

artem-zinnatullin:
cc @karlicoss

nikitin-da:
dine - final too. But it'll brake api 😿

artem-zinnatullin:
What API? Sorry, I'm out of context..

nikitin-da:
We can't publish it with 1.10.1 or 1.11.0 versions, becasue someone perharbs already override these methods

artem-zinnatullin:
Oh, I see, they were protected, then of course there is no sense to make them `final`… My bad!

Let's just make them public then :)

nikitin-da:
Okay =)

rsinukov:
ok

PR #698, opened 2016-10-21
Add to RawQuery affectsTables and observesTables that take collection

### LINKED ISSUE #696: Add RawQuery.CompleteBuilder.observesTables(Set<String>) method

nikitin-da:
Closes #696


artem-zinnatullin:
Hm, not sure I like that you can call both new and old api and it'll clear previous state…

nikitin-da:
I think adding to existing collection will be more complicated=(
And yet another level of builder that will protect calling both methods in compile time looks 
strangely

artem-zinnatullin:
Ok, let's add this variant, I've checked other libs with multiple methods of setting same value, they all rewrite result of previos call

PR #699, opened 2016-10-21
Queries take generic args instead of objects

### LINKED ISSUE #697: RawQuery.CompleteBuilder.args should be generic instead of taking vararg of objects

nikitin-da:
Closes #697


artem-zinnatullin:
👍

PR #720, opened 2016-11-19
#528: Support for AutoValue in StorIOSQLiteProcessor

geralt-encore:
https://github.com/pushtorefresh/storio/issues/528
https://github.com/pushtorefresh/storio/issues/614

I would appreciate help with testing, refactoring and overall logic improvement since it's quite a big feature and it required a lot of changes.

Issue #622, opened 2016-03-01
Sample Application is too complicated


careful7j:
Please, create a better (easier) sample application. Sample application should not rely on 25+ libraries and 9! projects. Thats cool that you know all of that, but some people may not. Sample app is very huge and it looks like some methods are already outdated.

If you want many people use the wonderful technology you designed, **it should be easy to start with**. Not everybody have 5+ years of expereince.

Rainer-Lang:
What do you mean with "9! projects"?

careful7j:
// Parts of StorIO
        storIOCommon                              : project(':storio-common'),
        storIOSQLite                              : project(':storio-sqlite'),
        storIOContentResolver                     : project(':storio-content-resolver'),
        storIOSQLiteAnnotations                   : project(':storio-sqlite-annotations'),
        storIOContentResolverAnnotations          : project(':storio-content-resolver-annotations'),
        storIOCommonAnnotationsProcessor          : project(':storio-common-annotations-processor'),
        storIOSQLiteAnnotationsProcessor          : project(':storio-sqlite-annotations-processor'),
        storIOContentResolverAnnotationsProcessor : project(':storio-content-resolver-annotations-processor'),
        storIOTestCommon                          : project(':storio-test-common'),

That one

Rainer-Lang:
That's the StorIO lib or better the parts of it. 
Later you will include only the parts you need eg contentresolver, sqlite,...

careful7j:
Well, sample app contains all of that things.

Rainer-Lang:
When did you start developing for android?

careful7j:
At least one year already passed (or mayby more).

Rainer-Lang:
My suggestion is - look at this parts which are really interesting for you right now, play with them and then look at the rest. 
I changed the sample app for me for testing. So I made my first steps. First,  I'm honest, it was hard for me - but I wanted a db-lib with rxjava - and storio was the right choice. Do you know rxjava? 

Focus on what you need, run the sample app and modify it.

careful7j:
Yeap, i have already tried rxjava with retrofit, its quite awesome. Now i am studying database solutions, i have tried ormlite and realm and everything was fine there. However, here I am stuck. Mayby storio is not something I can take with my level of expereince (after a whole night spent I am starting to think so).

Rainer-Lang:
Cool. Yes retrofit and rxjava are great. Storio is the perfect "addon". 😉
[...]

PR #722, opened 2016-11-27
kotlin data classes support in annotation processor

pbochenski:
I've added support for kotlin data classes. 
Please see kotlin sample app for usage and discussion of different approaches.


nikitin-da:
Hi, @pbochenski! 
Thank you for the work, but @geralt-encore has made this feature in #720 
The idea with sample application on kotlin is great! But will it be more united if it will be particular activity (package) in storio-sample-app? (To prevent multiple modules for each use case). 
And after @geralt-encore changes will be merged.
Thank you!

pbochenski:
Ah, I didn't see his PR. I just saw open issue and started working on it. I guess I was too slow 😞 
I can work on kotlin activity in sample app after @geralt-encore pr is merged :)

artem-zinnatullin:
@pbochenski damn, very sorry for that situation…

We've just merged #720, can you please check it with your Kotlin sample (only StorIOSQLite for now) and post a PR with sample app and changes required for it to work with changes made in #720? That would be very appreciated!

Thank you for contribution :)

pbochenski:
About the sample app. I think that doing separate app for kotlin is good idea and the reason is that kapt is needed for annotation processing in kotlin. And storeIO probably needs java example for people to look at. 
What do you think?

geralt-encore:
@pbochenski if you have any questions about implementation while doing sample app - just ask :)

nikitin-da:
Can we discuss it in issue #622 )?

PR #729, opened 2016-12-07
Release 1.12.0

nikitin-da:


CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
 StorIO Change Log
 ==========
 
+## Version 1.12.0
+
+_2016_12_7_
+
+* Support for **AutoValue** and **Kotlin data classes** in StorIOSQLiteProcessor and StorIOContentResolverProcessor. Many thanks to [@geralt-encore](https://github.com/geralt-encore)!
```
geralt-encore: It's not only about data classes. It can be used with a regular Kotlin class too.

nikitin-da: I'm sorry if I'm wrong, actually I still don't use kotlin at work=( 
Isn't processor work before for regular kotlin classes with maybe `lateinit` modifier?

geralt-encore: No, it didn't. The problem was that in Kotlin you don't work with directly with fields - you work with properties. And you couldn't apply column annotation on backed field property because it's private, which was limitation of the previous annotation processor implementation.

nikitin-da: Ok, thanks for clarification!

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
 StorIO Change Log
 ==========
 
+## Version 1.12.0
+
+_2016_12_7_
+
+* Support for **AutoValue** and **Kotlin** in StorIOSQLiteProcessor and StorIOContentResolverProcessor. Many thanks to [@geralt-encore](https://github.com/geralt-encore)!
```
artem-zinnatullin: 100500 thanks to be precise :)

CHANGELOG.md
```diff
@@ -1,6 +1,24 @@
...
+* DefaultStorIOSQLite combines affected tables from pending changes. **After transaction end will be send single notification, instead of multiple calls on every change**.
```
artem-zinnatullin: "now combines".  "After the end of transaction DefaultStorIOSQLite will send only one notification instead of multiple for every change." This should not be breaking behavior since `get()` operation is ready to consume such combined changes events.

CHANGELOG.md
artem-zinnatullin: Let's also add :party: emoji :)

nikitin-da:
Good luck on the conference, @artem-zinnatullin!

artem-zinnatullin:
:+1:

PR #739, opened 2016-12-17
Fix AutoValue integration

geralt-encore:


storio-common-annotations-processor/src/test/java/com/pushtorefresh/storio/common/annotations/processor/introspection/AnnotatedFieldValidationTest.java
```diff
@@ -30,15 +30,15 @@ public void failIfEnclosingElementIsNotType() {
...
-    @Test
-    public void failIfThereIsNoAnnotation() {
-        AnnotationProcessorStub stub = AnnotationProcessorStub.newInstance();
-        when(stub.enclosingElement.getAnnotation(AnnotationProcessorStub.TestClassAnnotation.class)).thenReturn(null);
-
-        expectedException.expect(ProcessingException.class);
-        expectedException.expectMessage("Please annotate class TestClass with TestClassAnnotation");
-        stub.processor.validateAnnotatedFieldOrMethod(stub.field);
-    }
+//    @Test
```
nikitin-da: To make test work we can mock superclass filed in `AnnotationProcessorStub` constructor:
```java
        TypeMirror fieldSuperClass = mock(TypeMirror.class);
        when(enclosingElement.getSuperclass()).thenReturn(fieldSuperClass);

        ProcessingEnvironment processingEnvironment = mock(ProcessingEnvironment.class);
        Types types = mock(Types.class);
        when(types.asElement(fieldSuperClass)).thenReturn(mock(Element.class));
        when(processingEnvironment.getTypeUtils()).thenReturn(types);
        doCallRealMethod().when(processor).init(processingEnvironment);
        processor.init(processingEnvironment);
```
Test with so much mocking are cut off from reality, but better than nothing =(

geralt-encore: I really like how annotation processor tests are organised in [Epoxy](https://github.com/google/compile-testing) using [Google Compile Testing](https://github.com/airbnb/epoxy). I tried to set up it in StorIO couple of months ago but failed. Maybe it requires more investigation.

nikitin-da:
@artem-zinnatullin check please, then we can merge it)

artem-zinnatullin:
@nikitin-da https://github.com/blog/2291-introducing-review-requests

nikitin-da:
Okay

storio-common-annotations-processor/src/test/java/com/pushtorefresh/storio/common/annotations/processor/introspection/AnnotatedFieldValidationTest.java
artem-zinnatullin: Lois

storio-common-annotations-processor/src/main/java/com/pushtorefresh/storio/common/annotations/processor/StorIOAnnotationsProcessor.java
```diff
@@ -98,10 +100,13 @@ private TypeElement validateAnnotatedClass(@NotNull final Element annotatedEleme
      * Checks that element annotated with {@link StorIOColumnMeta} satisfies all required conditions.
      *
      * @param annotatedElement an annotated field
+     * @return boolean flag which indicates if annotated element valid or not
      */
-    protected void validateAnnotatedFieldOrMethod(@NotNull final Element annotatedElement) {
+    protected boolean validateAnnotatedFieldOrMethod(@NotNull final Element annotatedElement) {
```
artem-zinnatullin: I would prefer to rename this method or throw specific exception, because at the moment this code mixes two concepts `return value` and `exceptions` that define how to behave for the caller, either we need to convert everything to return values (ie enum instead of exceptions) or throw exception instead of return value

geralt-encore: I mean I can change to a checked exception, but I think it will be even worse than the current implementation. We can't recover from that state and we don't need to print anything for user, so it will be just an empty catch block.

artem-zinnatullin: Well, what if exception will have very straingforward name? :) `SkipBecauseParentIsAnnotated`

artem-zinnatullin: Well, what if exception will have very straingforward name? :) `SkipBecauseParentIsAnnotated`

storio-sqlite-annotations-processor/src/main/java/com/pushtorefresh/storio/sqlite/annotations/processor/StorIOSQLiteProcessor.java
```diff
@@ -95,41 +96,46 @@ protected void processAnnotatedFieldsOrMethods(@NotNull final RoundEnvironment r
...
+                if (!storIOSQLiteTypeMeta.needCreator && storIOSQLiteColumnMeta.isMethod()) {
+                    storIOSQLiteTypeMeta.needCreator = true;
+                }
 
-            // Put meta column info.
-            storIOSQLiteTypeMeta.columns.put(storIOSQLiteColumnMeta.storIOColumn.name(), storIOSQLiteColumnMeta);
+                // Put meta column info.
+                storIOSQLiteTypeMeta.columns.put(storIOSQLiteColumnMeta.storIOColumn.name(), storIOSQLiteColumnMeta);
+            } catch (SkipNotAnnotatedClassWithAnnotatedParentException e) {
+                e.printStackTrace();
```
artem-zinnatullin: mm, do we need this?

geralt-encore: you mean `printStackTrace`? Not really, I can just delete it and leave the block empty

artem-zinnatullin: I guess that would be better :) What about logging it via annotation processor logger (forgot its name) with warning/etc level?

nikitin-da:
Thank you @geralt-encore!

PR #743, opened 2016-12-28
Update readme with kapt2 for Kotlin support

geralt-encore:
I would strongly suggest making a release as soon as possible since both AutoValue and Kotlin support don't work atm, but were advertised as the main features of the previous release.
And also it would be nice to merge this https://github.com/pushtorefresh/storio/pull/742 PR because it is on point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -210,7 +210,13 @@ public abstract class Tweet {
 }
 ```
 
-`Kotlin`
+`Kotlin`:
+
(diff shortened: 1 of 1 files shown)


artem-zinnatullin:
Will do tonight!

artem-zinnatullin:
Sorry, returned too late, will do release in the morning.

PR #747, opened 2017-01-01
Updated gradle/plugin/tools versions

nikitin-da:
Changelog:
https://developer.android.com/studio/releases/gradle-plugin.html
https://docs.gradle.org/current/release-notes
https://developer.android.com/studio/releases/build-tools.html


artem-zinnatullin:
Build failed :(

Issue #663, opened 2016-06-15
StorIOSQLite: add support for notification tags


nikitin-da:
Add the same thing like [`affectsTables`](https://github.com/pushtorefresh/storio/blob/c8c4f8a58888c97594acef011722f86fed9a0d11/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/queries/RawQuery.java) in all query classes, to make it possible to emit notifications of some 'uris', real tables for which aren't exist.

`storIOSQLite.observeChangesInTable(TAG)`

`InsertQuery.builder().table(TABLE).tags(TAG).build()`

@artem-zinnatullin, @thevery will it make sense or variant with [manual `notifyAboutChanges` call](https://github.com/pushtorefresh/storio/compare/dummy-table-changes?expand=1) is also applicable?

artem-zinnatullin:
I'd start without API for manual notifyAboutChanges because DB operations
should be single source of true for such system. Let's see how users will
use this feature and if somebody will request manual API.

Rainer-Lang:
I don't know if I'm right here....
I have a bunch of raw-sqls which affect different tables (I really don't know which exactly).
After executing these sqls I want to tell manually a defined set of "affected tables".
Is this now possible?

nikitin-da:
Can't you call something like this?
```java
storIOSQLite
   .executeSQL()
   .withQuery(RawQuery.builder()
      .query("delete from users")
      .affectsTables(users) // table names here
      .build())
   .prepare()
   .executeAsBlocking();
```
Or what do you mean by `don't know which exactly`?

Rainer-Lang:
Sorry, I should have provided more information.

I get a bunch of sqls from a server (during a sync process). I really don't now which tables are affected exactly.
I execute these sqls via ```executeSQL()```.
After processing all these sqls I want to to make a "manually affectedTables(syncTables)" at the end (not after every ```executeSQL()```).

worklflow
1) get sqls from server
2) ```executeSQL()``` every sql from 1)
3) at the end of 2): manual affectedTables(syncTables)

additional question:
Is it possible to give StorIO not only one sql, but a list of sqls to execute?

nikitin-da:
When you call `executeSQL()` `DefaultStorIOSQLite` uses `sqlite#execSQL` and according to documentation for it:  `Multiple statements separated by semicolons are not supported.` =(

But you can generate the list of `PreparedOperation` and execute it inside transaction.
```java
[...]

PR #768, opened 2017-02-26
Add notification tags

nikitin-da:
Closes: #663

### DIFF
--- a/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
+++ b/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
@@ -1,11 +1,13 @@
 package com.pushtorefresh.storio.sqlite;
 
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 
 import java.util.Collections;
 import java.util.Set;
 
 import static com.pushtorefresh.storio.internal.Checks.checkNotNull;
+import static java.util.Collections.singleton;
 
 /**
  * Immutable container of information about one or more changes happened in {@link StorIOSQLite}.
@@ -19,13 +21,37 @@ public final class Changes {
     private final Set<String> affectedTables;
 
     /**
+     * Immutable set of affected tags.
+     */
+    @NonNull
+    private final Set<String> affectedTags;
+
+    /**
      * Creates {@link Changes} container with info about changes.
      *
      * @param affectedTables set of tables which were affected by these changes.
+     * @param affectedTags set of tags which were affected by these changes.
      */
-    private Changes(@NonNull Set<String> affectedTables) {
+    private Changes(@NonNull Set<String> affectedTables, @NonNull Set<String> affectedTags) {
         checkNotNull(affectedTables, "Please specify affected tables");
+        checkNotNull(affectedTags, "Please specify affected tags");
         this.affectedTables = Collections.unmodifiableSet(affectedTables);
+        this.affectedTags = Collections.unmodifiableSet(affectedTags);
+    }
+
+    /**
+     * Creates new instance of {@link Changes}.
+     *
+     * @param affectedTables non-null set of affected tables.
+     * @param affectedTags non-null set of affected tags.
+     * @return new immutable instance of {@link Changes}.
+     */
+    @NonNull
+    public static Changes newInstance(
+            @NonNull Set<String> affectedTables,
+            @NonNull Set<String> affectedTags
+    ) {
+        return new Changes(affectedTables, affectedTags);
     }
 
     /**
@@ -36,7 +62,7 @@ public final class Changes {
      */
     @NonNull
     public static Changes newInstance(@NonNull Set<String> affectedTables) {
-        return new Changes(affectedTables);
+        return new Changes(affectedTables, Collections.<String>emptySet());
     }
 
     /**
@@ -47,8 +73,23 @@ public final class Changes {
      */
     @NonNull
     public static Changes newInstance(@NonNull String affectedTable) {
+        return newInstance(affectedTable, null);
+    }
+
+    /**
+     * Creates {@link Changes} container with info about changes.
+     *
+     * @param affectedTable table that was affected.
+     * @param affectedTag tag that was affected.
+     * @return new immutable instance of {@link Changes}.
+     */
+    @NonNull
+    public static Changes newInstance(@NonNull String affectedTable, @Nullable String affectedTag) {
         checkNotNull(affectedTable, "Please specify affected table");
-        return new Changes(Collections.singleton(affectedTable));
+        final Set<String> tags = affectedTag == null
+                ? Collections.<String>emptySet()
+                : singleton(affectedTag);
+        return new Changes(Collections.singleton(affectedTable), tags);
     }
 
     /**
@@ -61,6 +102,16 @@ public final class Changes {
         return affectedTables;
     }
 
+    /**
+     * Gets immutable set of affected tags.
+     *
+     * @return immutable set of affected tags.
+     */
+    @NonNull
+    public Set<String> affectedTags() {
+        return affectedTags;
+    }
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
@@ -68,18 +119,23 @@ public final class Changes {
 
         Changes changes = (Changes) o;
 
-        return affectedTables.equals(changes.affectedTables);
+        if (!affectedTables.equals(changes.affectedTables)) return false;
+        return affectedTags.equals(changes.affectedTags);
+
     }
 
     @Override
     public int hashCode() {
-        return affectedTables.hashCode();
+        int result = affectedTables.hashCode();
+        result = 31 * result + affectedTags.hashCode();
+        return result;
     }
 
     @Override
     public String toString() {
         return "Changes{" +
                 "affectedTables=" + affectedTables +
+                ", affectedTags=" + affectedTags +
                 '}';
     }
 }
--- a/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
+++ b/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
@@ -3,11 +3,14 @@ package com.pushtorefresh.storio.sqlite;
 import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
-import java.util.Collections;
+import java.util.Collection;
 import java.util.Set;
 
+import static com.pushtorefresh.storio.internal.Checks.checkNotEmpty;
 import static com.pushtorefresh.storio.internal.Checks.checkNotNull;
+import static com.pushtorefresh.storio.internal.InternalQueries.nonNullSet;
 import static java.util.Collections.singleton;
+import static java.util.Collections.unmodifiableSet;
 
 /**
  * Immutable container of information about one or more changes happened in {@link StorIOSQLite}.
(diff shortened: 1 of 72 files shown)


storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
```diff
@@ -36,7 +62,7 @@ private Changes(@NonNull Set<String> affectedTables) {
      */
     @NonNull
     public static Changes newInstance(@NonNull Set<String> affectedTables) {
-        return new Changes(affectedTables);
+        return new Changes(affectedTables, Collections.<String>emptySet());
```
artem-zinnatullin: omg, this looks so weird after almost year of writing in kotlin style `Collections.emptySet<String>()`. I haven't been really writing in java for a long time now…

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
```diff
@@ -61,25 +102,40 @@ public static Changes newInstance(@NonNull String affectedTable) {
...
+
     @Override
     public boolean equals(Object o) {
         if (this == o) return true;
         if (o == null || getClass() != o.getClass()) return false;
 
         Changes changes = (Changes) o;
 
-        return affectedTables.equals(changes.affectedTables);
+        if (!affectedTables.equals(changes.affectedTables)) return false;
```
artem-zinnatullin: Glad we have EqualsVerifier!

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
```diff
@@ -144,6 +145,110 @@
     }
 
     /**
+     * Allows observe changes of required tags.
+     * <p/>
+     * Tags are the addition level of notifications that provides
```
artem-zinnatullin: `Tags are optional meta information that you can attach to Changes object to have more fine-grained control over observing changes in the database.`?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
```diff
@@ -19,13 +21,37 @@
     private final Set<String> affectedTables;
 
     /**
+     * Immutable set of affected tags.
+     */
+    @NonNull
+    private final Set<String> affectedTags;
```
artem-zinnatullin: Btw, what if we make `Set<Object> affectedTags`? Or that would be overkill?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
```diff
@@ -144,6 +145,110 @@
...
+     * @param tag tag name to monitor.
+     * @return {@link rx.Observable} of {@link Changes} subscribed to changes of required tag.
+     */
+    @NonNull
+    public Observable<Changes> observeChangesOfTag(@NonNull String tag) {
+        checkNotNull(tag, "Tag can not be null or empty");
+        return observeChangesOfTags(Collections.singleton(tag));
+    }
+
+
```
artem-zinnatullin: el

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
```diff
@@ -144,6 +145,110 @@
...
+     * <p/>
+     * Notice, that returned {@link Observable} is "Hot Observable", it never ends, which means,
+     * that you should manually unsubscribe from it to prevent memory leak.
+     * Also, it can cause BackPressure problems.
+     *
+     * @param tags set of tags that should be monitored.
+     * @return {@link rx.Observable} of {@link Changes} subscribed to changes of required tags.
+     */
+    @NonNull
+    public abstract Observable<Changes> observeChangesOfTablesAndTags(
```
artem-zinnatullin: I don't think we need combination methods, user can observe raw changes and filter whatever combination he/she needs (using RxJava, plain java, etc)?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
```diff
@@ -144,6 +145,110 @@
...
+     * <p/>
+     * Notice, that returned {@link Observable} is "Hot Observable", it never ends, which means,
+     * that you should manually unsubscribe from it to prevent memory leak.
+     * Also, it can cause BackPressure problems.
+     *
+     * @param tag tag name to monitor.
+     * @return {@link rx.Observable} of {@link Changes} subscribed to changes of required tag.
+     */
+    @NonNull
+    public Observable<Changes> observeChangesOfTableAndTag(@NonNull String table, @NonNull String tag) {
```
artem-zinnatullin: Same as above?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/queries/UpdateQuery.java
```diff
@@ -222,6 +249,21 @@ public CompleteBuilder where(@Nullable String where) {
...
+         * Optional: Specifies notification tag to provide detailed information
+         * about which particular change were occurred.
+         *
+         * @param tag nullable notification tag name.
+         * @return builder.
+         * @see UpdateQuery#tag()
+         * @see com.pushtorefresh.storio.sqlite.StorIOSQLite#observeChangesOfTag(String)
+         */
+        @NonNull
+        public CompleteBuilder tag(@Nullable String tag) {
```
artem-zinnatullin: tags? or overload with `Set<String>`? Same for other queries

storio-sqlite/src/test/java/com/pushtorefresh/storio/sqlite/queries/RawQueryTest.java
```diff
@@ -65,6 +65,16 @@ public void observesTablesShouldNotBeNull() {
     }
 
     @Test
+    public void observesTagsShouldNotBeNull() {
+        RawQuery rawQuery = RawQuery.builder()
+                .query("lalala I know SQL")
```
artem-zinnatullin: kek

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
nikitin-da: I thought I will be less convenient for the most frequent case with `Strings`. 
But I agree tags it is not about common cases, so flexibility should be the main purpose.
@rsinukov also votes for `object`.
Let's do it=)

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
nikitin-da: :crying_cat_face:

storio-sqlite/src/test/java/com/pushtorefresh/storio/sqlite/queries/RawQueryTest.java
nikitin-da: nobody knows!

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
nikitin-da: If so, we should open `ChangesFilter` to make it visible for `Query`

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/queries/UpdateQuery.java
nikitin-da: 👍

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
nikitin-da: Lets make it `public` and keep comment that it is for internal usage only?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
artem-zinnatullin: Well, I do agree that Strings will probably be most used and we'll have to force these users to class-cast them for no good reason, let's wait for @geralt-encore opinion then!

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
artem-zinnatullin: I'd prefer to keep as small amount of public methods/classes/API as possible if current ones already allow you to do what you want without too much boilerplate…

storio-sqlite/src/test/java/com/pushtorefresh/storio/sqlite/queries/RawQueryTest.java
artem-zinnatullin: I won't tell @thevery and Stasic! (If you see this message it's a GitHub's bug obviously).

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
geralt-encore: +1 for `Object`. In my understanding, you can use as tag whatever you want and if `String` will be the most frequently used it doesn't mean that tag should be restricted to it. Also, `View` tag in Android is `Object` as well.

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/Changes.java
```diff
@@ -47,8 +73,23 @@ public static Changes newInstance(@NonNull Set<String> affectedTables) {
...
+     * @param affectedTag tag that was affected.
+     * @return new immutable instance of {@link Changes}.
+     */
+    @NonNull
+    public static Changes newInstance(@NonNull String affectedTable, @Nullable String affectedTag) {
         checkNotNull(affectedTable, "Please specify affected table");
-        return new Changes(Collections.singleton(affectedTable));
+        final Set<String> tags = affectedTag == null
+                ? Collections.<String>emptySet()
+                : singleton(affectedTag);
```
geralt-encore: It makes sense to reuse static import for `singleton` next line or remove it completely.

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/StorIOSQLite.java
geralt-encore: +1 for removing it. It can get out of hands pretty quick if we need to introduce one more way of filtering changes. `observeChangesOfTablesAndTagsAndWhatever` plus combination look ugly.

geralt-encore:
@nikitin-da  just one comment. I think that it might be a good idea to change signature of `varargs` methods which are part of public API to `T firstTag, T... tags` instead of just `T...tags` since with current implementation it is possible to forget to provide any tags which lead to very obscure bugs for users.

nikitin-da:
@geralt-encore good idea! 
For queries it will be quite simple (if someone will decided to reuse existed builder and clear tags, he can call `tags((Collection<?>) null)`).
But for changes and operation results it will increase overloaded methods =(
We can't remove methods with signature like `newInstance(@NonNull Set<String> affectedTables)` in minor release, because it is breaking change.
And creating separate method will increase class complexity (PutResult now 8 methods  ->will be 12 😿 )
@artem-zinnatullin, what is your opinion?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/execute/PreparedExecuteSQL.java
```diff
@@ -113,7 +118,22 @@ public Object executeAsBlocking() {
...
     public Observable<Object> asRxObservable() {
-        return RxJavaUtils.createObservable(storIOSQLite, this);
+        throwExceptionIfRxJavaIsNotAvailable("asRxObservable()");
+
+        final Set<String> observesTables = rawQuery.observesTables();
+        final Set<?> observesTags = rawQuery.observesTags();
+
+        final Observable<Object> observable;
+        if (!observesTables.isEmpty() || !observesTags.isEmpty()) {
+            observable = ChangesFilter.applyForTablesAndTags(storIOSQLite.observeChanges(), observesTables, observesTags)
```
nikitin-da: @artem-zinnatullin It seems we forgot to subscribe on changes here =(

nikitin-da:
@artem-zinnatullin PTAL please, to prevent guys from merge conflicts ;)

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/ChangesFilter.java
```diff
@@ -16,28 +17,62 @@
  * <p>
  * Hides RxJava from ClassLoader via separate class.
  */
-final class ChangesFilter implements Func1<Changes, Boolean> {
+public final class ChangesFilter implements Func1<Changes, Boolean> {
```
artem-zinnatullin: We can't keep it package private?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/execute/PreparedExecuteSQL.java
artem-zinnatullin: Please correct me if I'm wrong, but it was our design decision, since `executeSql` is basically a side effect and does not return data from database, same as we do for Put and Delete, we don't subscribe them to db changes!

thevery:
@artem-zinnatullin objects are easier to fail (`hashCode`/`equals` override) and I don't know useful examples of objects instead of strings, especially with the possibility to pass several tags.

artem-zinnatullin:
Yeah, I have similar thoughts, moreover, string tags can be more understandable for those who remember how web 2.0 was growing: tag clouds on websites, tags in blog posts/articles, tags in twitter and so on. @geralt-encore?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/execute/PreparedExecuteSQL.java
nikitin-da: Yep, I agree that it is the side effect operation and it will not be clean to subscribe on some changes with it. 
But current `RawQuery.observesTags` possibility can be misleading.
Maybe we should note this feature in doc, because it may look as unexpected behavior =(

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/ChangesFilter.java
nikitin-da: No, because it used in `com.pushtorefresh.storio.sqlite.impl` for `DefaultStorIOSQLite#observeChangesInTables` and `com.pushtorefresh.storio.sqlite.operations.execute` for `asRxObservable`

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/execute/PreparedExecuteSQL.java
artem-zinnatullin: Yes, I definitely remember discussing that, and I'm pretty sure we have docs about it :)

nikitin-da:
It seem there is equality of votes for now, so what type of tag will be final?) 
@geralt-encore, @rsinukov, @stanisvs voted for objects 👍 , 
@artem-zinnatullin and @thevery and I - strings 👎

nikitin-da:
There was also question about first required tag with vararg https://github.com/pushtorefresh/storio/pull/768#issuecomment-283715392
@artem-zinnatullin what is your opinion?

storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/execute/PreparedExecuteSQL.java
nikitin-da: I didn't find it for [asRxObservable](https://github.com/pushtorefresh/storio/blob/master/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/operations/execute/PreparedExecuteSQL.java#L68), for [RawQuery](https://github.com/pushtorefresh/storio/blob/master/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/queries/RawQuery.java#L270) and in the [doc](https://github.com/pushtorefresh/storio/blob/master/docs/StorIOSQLite.md#4-execsql-operation) too=)

artem-zinnatullin:
@nikitin-da I've changed emojis to ❤️ and 🎉 to remove negativity of 👎 variant haha, so please ask guys from your team to vote here so we could count here, cc @geralt-encore, I'll also ask community to vote. 

Regarding overloads — I like @geralt-encore's suggestion about `T tag, T… otherTags`, so if that seems possible now (since we decided to remove `observeChangesOfTablesAndTags()` — let's implement that for methods that accept tags only. I didn't really get the note about removing`newInstance(@NonNull Set<String> affectedTables)`, sorry…

nikitin-da:
String variant wont 6:5 )

PR #770, opened 2017-03-07
Dependencies

geralt-encore:


geralt-encore:
hm I don't have anymore  lint errors if I run it locally 🤔

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.2.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
```
artem-zinnatullin: I don't see gradle wrapper files updated, please run something like `./gradlew --version` and commit downloaded files!

geralt-encore: `./gradlew --version` didn't work, but `gradle wrapper` did the trick

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sun Jan 01 23:39:39 YEKT 2017
+#Mon Mar 13 23:28:38 EET 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.9-bin.zip
```
artem-zinnatullin: Wrong version!

geralt-encore: ouch

artem-zinnatullin:
You can `cat storio-basic-sample-app/build/reports/lint-results.xml` after build to see what problem lint detect on CI that you don't see locally

geralt-encore:
You mean run it on Travis? Probably I don't have sufficient rights to do it. At least I don't see how to do it. The main question is why result on CI differs from a local one. How is it even possible? I mean, I understand if there are some configuration issues or different environment. But here we have additional lint issue, which can be found on the same project built on different machine.

geralt-encore:
```
<issues format="4" by="lint 2.3.0">
    <issue
        id="NewApi"
        severity="Error"
        message="The SDK platform-tools version (23.0.1) is too old  to check APIs compiled with API 25; please update"
        category="Correctness"
        priority="6"
        summary="Calling new methods on older versions"
        explanation="This check scans through all the Android API calls in the application and warns about any calls that are not available on *all* versions targeted by this application (according to its minimum SDK attribute in the manifest).&#xA;&#xA;If you really want to use this API and don&apos;t need to support older devices just set the `minSdkVersion` in your `build.gradle` or `AndroidManifest.xml` files.&#xA;&#xA;If your code is *deliberately* accessing newer APIs, and you have ensured (e.g. with conditional execution) that this code will only ever be called on a supported platform, then you can annotate your class or method with the `@TargetApi` annotation specifying the local minimum SDK to apply, such as `@TargetApi(11)`, such that this check considers 11 rather than your manifest file&apos;s minimum SDK as the required API level.&#xA;&#xA;If you are deliberately setting `android:` attributes in style definitions, make sure you place this in a `values-vNN` folder in order to avoid running into runtime conflicts on certain devices where manufacturers have added custom attributes whose ids conflict with the new ones on later platforms.&#xA;&#xA;Similarly, you can use tools:targetApi=&quot;11&quot; in an XML file to indicate that the element will only be inflated in an adequate context."
        quickfix="studio">
        <location
            file="/home/travis/build/pushtorefresh/storio/storio-basic-sample-app"/>
    </issue>
</issues>
```
But I have no idea where exactly it is located

.travis.yml
```diff
@@ -22,6 +22,9 @@ script:
 after_success:
   - codecov
 
+after_failure:
+  - cat storio-basic-sample-app/build/reports/lint-results.xml
```
artem-zinnatullin: I think we can remove it now :)

artem-zinnatullin:
Looks like Travis does not download latest platform-tools, we can try to download them manually like this:

```bash
export ANDROID_SDK_INSTALL_COMPONENT="echo \"y\" | \"$ANDROID_HOME\"/tools/android update sdk --no-ui --all --filter"

"$ANDROID_HOME"/tools/android list sdk --all --no-ui --extended

eval $ANDROID_SDK_INSTALL_COMPONENT "platform-tools"
```

geralt-encore:
I doubt it. I've seen plenty of projects working fine with `android-25` on Travis. Should be something different, but I can't even imagine what it could be.

artem-zinnatullin:
Well, this is bad… Can we update everything except android-25 (and target api) for now?

geralt-encore:
Ok, I'll just close this one and will make a separate PR with Gradle update. Not sure that updating dependencies for samples actually worth it.

Issue #735, opened 2016-12-12
Auto generated GetResolver has constructor with wrong parameter order in Kotlin


pbochenski:
Reproduction steps: 
1. create class with more than two columns f.e.
```kotlin
@StorIOSQLiteType(table = "CreditCard")
data class CreditCard @StorIOSQLiteCreator constructor(
        @get:StorIOSQLiteColumn(name = "id", key = true) val cardId: Long,
        @get:StorIOSQLiteColumn(name = "name") val name: String,
        @get:StorIOSQLiteColumn(name = "expireYear") val expireYear: Int,
        @get:StorIOSQLiteColumn(name = "expireMonth") val expireMonth: Int,
        @get:StorIOSQLiteColumn(name = "type") val type: String,
        @get:StorIOSQLiteColumn(name = "isDefault") val isDefault: Boolean)
```
2. build project (couple of times if needed). 

The reason for that is kapt (kotlins annotation processor, don't know about apt) does not process annotations in creation order.

geralt-encore:
Did you check it with kapt or kapt2?

pbochenski:
kapt1 I guess (I'm using '1.0.5-2' version of Kotlin)   
the one with 
```
kapt {
    generateStubs = true
}
```

pbochenski:
just checked with the new kapt (the one described here: https://blog.jetbrains.com/kotlin/2016/09/kotlin-1-0-4-is-here/) and the problem is still occurring.

geralt-encore:
Looks like I was very (un)fortunate with my Kotlin test case, but it's still my fault because I haven't tested it properly and was more concentrated on AutoValue case. I asked a question in Kotlin slack if it is expected behaviour or not. If it works as expected then I'll add an ordering parameter to column annotations and will sort fields according to it.
Thank you, very much for filing this issue! And sorry for inconvenience =(

pbochenski:
maybe some reflection can be used to avoid parameter numbering?

geralt-encore:
JetBrains guys said that kapt should behave in the same way as apt.
So here is an issue: https://youtrack.jetbrains.com/issue/KT-15202.
Since we don't know when it will be fixed and current version doesn't work I think we still need to do something even if the problem is on kapt side.

geralt-encore:
@artem-zinnatullin Kotlin support is broken atm. Shall we wait more answers and fixes from Kotlin team or shall we start acting upon it by ourselves?

[...]

PR #776, opened 2017-04-11
Support for Kotlin properties

geralt-encore:
With this PR it is possible to annotate Kotlin properties directly utilising getters and setter instead of direct access to fields.
Also, it resolves https://github.com/pushtorefresh/storio/issues/735 since there is no need anymore to map fields to constructor parameters.

### DIFF
--- a/storio-common-annotations-processor/src/main/kotlin/com/pushtorefresh/storio/common/annotations/processor/StorIOAnnotationsProcessor.kt
+++ b/storio-common-annotations-processor/src/main/kotlin/com/pushtorefresh/storio/common/annotations/processor/StorIOAnnotationsProcessor.kt
@@ -7,6 +7,7 @@ import com.pushtorefresh.storio.common.annotations.processor.introspection.StorI
 import javax.annotation.processing.*
 import javax.lang.model.SourceVersion
 import javax.lang.model.element.Element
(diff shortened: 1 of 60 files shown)


storio-common-annotations-processor/src/main/kotlin/com/pushtorefresh/storio/common/annotations/processor/StorIOAnnotationsProcessor.kt
```diff
@@ -141,6 +152,53 @@ abstract class StorIOAnnotationsProcessor<TypeMeta : StorIOTypeMeta<*, *>, out C
...
+        var setter: String? = null
+        annotatedElement.enclosingElement.enclosedElements.forEach { element ->
+            if (element.kind == ElementKind.METHOD) {
+                val method = element as ExecutableElement
+                val methodName = method.simpleName.toString()
+                // check if it is a valid getter
+                if ((methodName == String.format("get%s", name.capitalize())
+                        || methodName == String.format("is%s", name.capitalize())
+                        // Special case for properties which name starts with is.
+                        // Kotlin will generate getter with the same name instead of isIsProperty.
```
nikitin-da: :neutral_face:

Issue #766, opened 2017-02-19
Release automation


artem-zinnatullin:
Currently, I'm the only person who can actually publish release to Maven Central. This is something I'm not proud of since it effectively reduces Bus Factor of the project to 1, even though we now have 3 persons with push rights to the repo: @nikitin-da, @geralt-encore, me.

- [ ] As a first step I'd like to find a way to give @nikitin-da access and setup his environment so he could also do manual release. If that'll work without need to give access to my personal sonatype account, that'll be awesome.
- [ ] As a second step I'd like to setup release automation on CI with signed tags and ecrypted keys so any member of the project could do a release by simply pushing signed tag of a new version.

@nikitin-da, @geralt-encore thoughts?

nikitin-da:
Let's do it! 👍 
It will allow you not to be distracted by some less important things and sometimes even to sleep! (Without clones creating like it probably does Jake).
Release via tags sounds adorably=)

geralt-encore:
I am always for automating as much thing is possible, so let me know if you need any help with it 👍

artem-zinnatullin:
So I've checked the options, when we'll have something to release, I'll try to do it via https://github.com/Codearte/gradle-nexus-staging-plugin and if it'll work — I'll make a PR so we could do it automatically.

artem-zinnatullin:
I've checked it on another project and it works! Will also require some travis setup https://docs.travis-ci.com/user/deployment/releases/

thevery:
@artem-zinnatullin some thoughts on the subj:
1) What about jcenter instead of sonatype? At least for releases.
2) We know you're a fan of tags instead of branches policy, but I still like auto-publish on every commit in release branch

artem-zinnatullin:
1. "At least for releases?" but we don't publish snapshots, so it's only
releases. I was recently setting up publishing to jcenter in some Juno
projects and faced major problem issue with publishing package with several
artifacts which took Bintray a week to resolve and they still don't have
automated solution for it
https://stackoverflow.com/questions/43468385/bintray-does-not-sync-one-of-the-artifacts-of-the-package-to-the-jcenter

2. I don't see reasons for that, without tags you have to manually store
version in source code, update it and so on and so on, tags are great, take
a look at release automation here https://github.com/gojuno/composer (I'm
planning to do something very similar in StorIO)

PR #788, opened 2017-05-14
Add Gradle plugin for automatic Sonatype releases.

artem-zinnatullin:
This is the first step in Sonatype release automation #766.

Release can be done on machine with credentials for it by running `./gradlew clean uploadArchives --info closeAndReleaseRepository`.

Next step would be adding encrypted credentials to Travis and setting up builds scripts so that pushing git tag would trigger the release.

v1.13.0 was released using this plugin.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,6 +20,9 @@ buildscript {
         // Kotlin
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
 
+        // Automatic releases to Sonatype.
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -149,3 +152,8 @@ subprojects {
     }
 }
 
+apply plugin: 'io.codearte.nexus-staging'
+
+nexusStaging {
+    packageGroup = 'com.pushtorefresh'
```
artem-zinnatullin: Yeah, unfortunately it'll try to close and release all `com.pushtorefresh` projects that are in staging mode on Sonatype, luckily we don't do releases that often so clash is very unlikely.

PR #793, opened 2017-05-18
Update Android Gradle Plugin to 2.3.2, support libs to 25.3.1.

artem-zinnatullin:
I suspect it'll cause Travis Lint failure but I'll try to fix.


build.gradle
```diff
@@ -42,11 +42,11 @@ allprojects {
 
 ext {
     buildToolsVersion = '25.0.2'
-    compileSdkVersion = 23
+    compileSdkVersion = 25
     minSdkVersion = 14
 }
 
-def supportLibsVersion = '23.4.0'
+def supportLibsVersion = '25.3.1'
```
artem-zinnatullin: Apparently Lint from new Android Gradle Plugin detected class copy of the class that is present in new support libs, so I had to update support libs, but update of support libs required bumping compileSdkVersion.

Uh…

artem-zinnatullin:
closed to prevent accidental merge

PR #794, opened 2017-05-22
Update Android Gradle Plugin to 2.3.2, support libs to 25.3.1.

artem-zinnatullin:
Opened instead of #793.

I now manually download Android SDK which resolves wrong "platform-tools" version problem on Travis CI.


artem-zinnatullin:
Yeah, looks like so, never liked their magic dsl

artem-zinnatullin:
Codecov thinks that coverage will be reduced by `-0.01` which is about its rate of error, so I'm merging this

Issue #797, opened 2017-06-12
AnnotationProcessor fail with AutoValue 


Sar777:
Hi, I am using AutoValue and have this result: 

```
@AutoValue
@StorIOSQLiteType(table = CharactersTable.TABLE)
public abstract class LocalCharacter {

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_GUID, key = true)
    public abstract Long guid();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_NAME)
    public abstract String name();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_RACE)
    public abstract Integer race();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_CLASS)
    public abstract Integer clazz();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_GENDER)
    public abstract Integer gender();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_MONEY)
    public abstract Long money();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_TOTAL_KILLS)
    public abstract Integer totalKills();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_TODAY_KILLS)
    public abstract Integer todayKills();

    @NonNull
    @StorIOSQLiteColumn(name = CharactersTable.COLUMN_YESTERDAY_KILLS)
    public abstract Integer yesterdayKills();
    
    @StorIOSQLiteCreator
    static LocalCharacter create(@NonNull Long guid, @NonNull String name, @NonNull Integer race, @NonNull Integer clazz, @NonNull Integer gender, @NonNull Long money, @NonNull Integer totalKills, @NonNull Integer todayKills, @NonNull Integer yesterdaykills) {
        return new AutoValue_LocalCharacter(guid, name, race, clazz, gender, money, totalKills, todayKills, yesterdaykills);
    }
}
```
**Error:**
`Error:Problem occurred with StorIOProcessor: -1`

AutoValue: **1.4.1**
StorIO: **1.13.0**

Sorry my bad english. Thx.

geralt-encore:
Thanks for reporting! I'll look into it as soon as I can.

Sar777:
@geralt-encore thx.

geralt-encore:
[...]

PR #798, opened 2017-06-13
#797 Proper message in case creator parameters do not match columns

geralt-encore:
https://github.com/pushtorefresh/storio/issues/797


storio-content-resolver-annotations-processor-test/src/test/resources/CreatorWithNotMatchingArguments.java
```diff
@@ -0,0 +1,17 @@
...
+
+    private long id;
+
+    @StorIOContentResolverColumn(name = "id", key = true)
+    long id() {
+        return id;
+    }
+
+    @StorIOContentResolverCreator
+    public CreatorWithNotMatchingArguments(long it) {
```
nikitin-da: Maybe rename to "field" and "anotherField" or smth else.
Example in misprint style is fun but as for me not so notable=(

Issue #801, opened 2017-07-03
Kotlin Data Class support


mattdevelops:
Hello, I've recently dug up into StarIO and so far the API seemed nice and smooth until I tried applying it to a Kotlin Data Class. Is there any support for it, or am I using bad annotations here ? I keep getting this issue:
`Error:(19, 1) Some error(s) occurred while processing annotations. Please see the error messages above.`

And my data class looks like:

```
@StorIOSQLiteType(table = "users")
data class User @StorIOSQLiteCreator constructor(
        @StorIOSQLiteColumn(name = "_id", key = true) val _id: String = "",
        @StorIOSQLiteColumn(name = "email") val email: String,
        @StorIOSQLiteColumn(name = "password") val password: String? = null,
        @StorIOSQLiteColumn(name = "firstName") val firstName: String,
        @StorIOSQLiteColumn(name = "profilePicture") val profilePicture: String? = null,
        @StorIOSQLiteColumn(name = "location") val location: String? = null,
        @StorIOSQLiteColumn(name = "createdAt") val createdAt: String? = null,
        @StorIOSQLiteColumn(name = "updatedAt") val updatedAt: String? = null)
```

geralt-encore:
Hi! Are there any messages about an error or it is the only one? Because if it is the only one - it is not very helpful and we need to improve it for sure.
Regarding the error, you need to change properties declaration from `val` to `var` since we need both getter and setter to be able to generate correct code for this case.

mattdevelops:
Got it, yes the only error I'm getting is the one I provided, plus a warning:
`Warning:warning: The following options were not recognized by any processor: '[kapt.kotlin.generated]'` , besides that nothing more.


Here's the final code that seems to work now
```
@StorIOSQLiteType(table = "users")
data class User @StorIOSQLiteCreator constructor(
        @StorIOSQLiteColumn(name = "_id", key = true) var _id: String = "",
        @StorIOSQLiteColumn(name = "email") var email: String,
        @StorIOSQLiteColumn(name = "password") var password: String? = null,
        @StorIOSQLiteColumn(name = "firstName") var firstName: String,
        @StorIOSQLiteColumn(name = "profilePicture") var profilePicture: String? = null,
        @StorIOSQLiteColumn(name = "location") var location: String? = null,
        @StorIOSQLiteColumn(name = "createdAt") var createdAt: String? = null,
        @StorIOSQLiteColumn(name = "updatedAt") var updatedAt: String? = null) {

[...]

PR #802, opened 2017-07-06
#801 Fix Kotlin integration

geralt-encore:
https://github.com/pushtorefresh/storio/issues/801
Previous Kotlin integration was broken by design - it was forcing you to use `vars` instead of `vals` and adding an empty constructor. Dunno what I was thinking while doing it in a first place 😔 
Current one solves this issues. The only problem is that it is not backwards compatible with the previous one which didn't work as intended so I am not really sure how big is the problem and how many people are affected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -188,6 +188,20 @@ public class Tweet {
 }
 ```
 
+-`Kotlin`:
+
+In order to make annotation processors work with Kotlin you need to add the following to your `build.gradle`:
+```groovy
+apply plugin: 'kotlin-kapt'
+```
+Then use `kapt` configuration instead of `annotationProcessor`.
+
+ ```kotlin
+ @StorIOSQLiteType(table = "tweets") data class Tweet @StorIOSQLiteCreator constructor(StorIOSQLiteColumn(name = "author") var author: String,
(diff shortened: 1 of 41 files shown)


README.md
```diff
@@ -188,6 +188,20 @@ public class Tweet {
...
+-`Kotlin`:
+
+In order to make annotation processors work with Kotlin you need to add the following to your `build.gradle`:
+```groovy
+apply plugin: 'kotlin-kapt'
+```
+Then use `kapt` configuration instead of `annotationProcessor`.
+
+ ```kotlin
+ @StorIOSQLiteType(table = "tweets") data class Tweet @StorIOSQLiteCreator constructor(StorIOSQLiteColumn(name = "author") var author: String,
```
nikitin-da: May we format it like this, to fit it into screen
```kotlin
@StorIOSQLiteType(table = "tweets")
data class Tweet @StorIOSQLiteCreator constructor(
        StorIOSQLiteColumn(name = "author") var author: String,
        StorIOSQLiteColumn(name = "content") var content: String
)
```

geralt-encore:
Also, I think that we need some basic Kotlin sample to be able to do some basic validation against it. Testing annotation processor against Java is fine, but not enough.

PR #771, opened 2017-03-18
Interceptors api

rsinukov:
Hi.
There is a proof of concept for interceptors API and sample `LoggingInterceptor` to see how it works.
Main idea of this PR is to create `Interceptor.Chain` in every `PreparedOperation#executeAsBlocking()` method and put in chain `RealInterceptor`, which will do real job.  So, all code in `RealInterceptor#intercept()` is copypaste of past `executeAsBlocking()` code.

PR #803, opened 2017-07-10
Fix case with different classes having fields with the same names

geralt-encore:


README.md
```diff
@@ -189,6 +189,7 @@ public class Tweet {
 ```
 
 -`Kotlin`:
```
nikitin-da: excess dash

geralt-encore:
@artem-zinnatullin @nikitin-da how about making a release with Kotlin related changes? Seems reasonable considering the fact of its growing popularity.

nikitin-da:
@geralt-encore I agree that there is huge pack of changes and Kotlin changes required to release ASAP, but it seems RxJava2 support will break compatibility again. 
It will lead to two major releases at once =(

geralt-encore:
I see your point, but I have to concerns regarding it. First, it is not really a breaking API change - it is fixing something that was initially broken, so I think that the fact that something doesn't work in a broken way, which it was never supposed to work is a good thing. Second, we have zero ideas about the timeline of RxJava 2 support =(

nikitin-da:
Unfortunately we have already merged [interceptors](https://github.com/pushtorefresh/storio/pull/771) and we should roll it back from master or release it with major version(
@artem-zinnatullin will it be ok to release fast v2 and v3?

geralt-encore:
Ouch, I completely forgot about interceptors change... Then, yeah, you are totally right.

artem-zinnatullin:
@nikitin-da yes, I think we can release v2 and then v3, that's what semantic versioning is for :)

nikitin-da:
Ok, let's release v2

artem-zinnatullin:
See #806

PR #806, opened 2017-07-25
Automate release process.

artem-zinnatullin:
Alright, I think I've finally managed to automate release process for StorIO, it was hard and I'm not sure if it'll work right away, but it should.

I combined multiple solutions and existing plugins I've used before, we now should be able to publish release to maven central by simply pushing a tag. You can do it right on GitHub releases page. 

Very similar set of scripts works for [gojuno/composer](https://github.com/gojuno/composer), [gojuno/swarmer](https://github.com/gojuno/swarmer), [gojuno/koptional](https://github.com/gojuno/koptional) and so on. 

However in StorIO we use Sonatype's Maven Central instead of Bintray's Jcenter which in certain ways complicate things but in others makes it easier.

Travis CI now securely (relatively) stores credentials to sign and push release to Maven Central, that should increase our bus factor from 1 (me) to 3 (@nikitin-da, @geralt-encore and me).

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -2,6 +2,28 @@
 set -e
 
 # Please run it from root project directory
+
 # For some reason test for annotation processor are failing on a regular CI setup.
 # So we had to exclude test task for it from the main build process and execute it as a separate command.
-./gradlew clean build checkstyle -PdisablePreDex -x :storio-sqlite-annotations-processor-test:test -x :storio-content-resolver-annotations-processor-test:test && ./gradlew :storio-sqlite-annotations-processor-test:testDebugUnitTest && ./gradlew :storio-content-resolver-annotations-processor-test:testDebugUnitTest
+./gradlew clean build checkstyle -PdisablePreDex -x :storio-sqlite-annotations-processor-test:test -x :storio-content-resolver-annotations-processor-test:test
+./gradlew :storio-sqlite-annotations-processor-test:testDebugUnitTest
+./gradlew :storio-content-resolver-annotations-processor-test:testDebugUnitTest
+
+if git describe --exact-match --tags $(git log -n1 --pretty='%h') ; then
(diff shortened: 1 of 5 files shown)


ci.sh
```diff
@@ -2,6 +2,28 @@
...
 # So we had to exclude test task for it from the main build process and execute it as a separate command.
-./gradlew clean build checkstyle -PdisablePreDex -x :storio-sqlite-annotations-processor-test:test -x :storio-content-resolver-annotations-processor-test:test && ./gradlew :storio-sqlite-annotations-processor-test:testDebugUnitTest && ./gradlew :storio-content-resolver-annotations-processor-test:testDebugUnitTest
+./gradlew clean build checkstyle -PdisablePreDex -x :storio-sqlite-annotations-processor-test:test -x :storio-content-resolver-annotations-processor-test:test
+./gradlew :storio-sqlite-annotations-processor-test:testDebugUnitTest
+./gradlew :storio-content-resolver-annotations-processor-test:testDebugUnitTest
+
+if git describe --exact-match --tags $(git log -n1 --pretty='%h') ; then
+    echo "Git tag detected, launching release process..."
+
+    if [ -z "$GPG_SECRET_KEYS" ]; then
```
artem-zinnatullin: During code review we all need to make sure that nobody tries to print our environment variables, because that way we'll effectively leak our credentials and allow others to sign and upload jars from our account.

gradle.properties
```diff
@@ -1,5 +1,4 @@
 VERSION_CODE=29
```
nikitin-da: how will we manage `VERSION_CODE`?

artem-zinnatullin: So it's actually not required, let's try to release with same version code and if it'll work we'll just drop it later

Issue #519, opened 2015-09-15
All Resolver's methods should receive instance of StorIO


artem-zinnatullin:
At the moment, users have to make some hacks to make it work…

Rainer-Lang:
+1

Issue #457, opened 2015-07-22
Remove Query.whereArgs(List<?> whereArgs) from StorIOSQLite


artem-zinnatullin:
This was my mistake in the API, I guess, nobody used it because we have nice overload which is used in all other Queries `whereArgs(Object… whereArgs)`.

I think it's okay to remove this API in 1.1.0?

@nikitin-da, anybody?

nikitin-da:
I agree. But have you decided that it will not break compatibility?

artem-zinnatullin:
Unfortunately it will break compatibility, but I hope nobody uses it because there is `Object... whereArgs` overload.

Recently OkHttp denied body in PUT and DELETE requests, I didn't see so many issues about one change for a while…

I think it's critical to made this change ASAP while we don't have a lot of users.

nikitin-da:
Ok. Lets remove it.

artem-zinnatullin:
Reopening, this is incompatible API change, so we need to move this issue to v2.0.0

PR with revert — #468

vokilam:
@artem-zinnatullin, @nikitin-da, before removing `whereArgs(List<?> whereArgs)` consider more complex cases of constructing a query. In example below, additional expressions are appended to where clause based on condition. One would probably implement it like this:

    boolean includeHappyTeenagers = true;
    StringBuilder whereBuilder = new StringBuilder();
    List<Object> argsList = new ArrayList<>();

    // required part
    whereBuilder.append("age BETWEEN ? AND ?");
    argsList.add(60);
    argsList.add(90);

    // optional part
    if (includeHappyTeenagers) {
        whereBuilder.append(" OR age BETWEEN ? AND ? AND mood > ?");
        argsList.add(13);
        argsList.add(18);
        argsList.add(0.7);
    }

    Cursor c = db.get().cursor()
            .withQuery(Query.builder()
                            .table("people")
                            .where(whereBuilder.toString())
                            .whereArgs(argsList)
                            .build()
            )
            .prepare()
            .executeAsBlocking();

    c.getCount();
    c.close();

Without `whereArgs(List<?> whereArgs)` same result will be possible to achieve by converting to a plain array:

    whereArgs(argsList.toArray())
    whereArgs(argsList.toArray(new Object[argsList.size()]))

that doesn't look beautiful.
[...]

PR #817, opened 2017-08-20
Add storIOSqlite parameter to get resolver

nikitin-da:
Part of #519


geralt-encore:
👍 
@nikitin-da, I can finish ContentResolver part if you want.

nikitin-da:
@geralt-encore thank you, but I'v already started to work on it, will finish in few hours =)
There will be only one release blocker #457. What is your opinion, should we remove this method or add the same to all other queries?

geralt-encore:
@nikitin-da cool! I commented in #457

PR #824, opened 2017-10-06
Fix Travis log overflow, update Android SDK tools 25.2.3 → 3859397.

artem-zinnatullin:
Should reduce amount of logs, sometimes Travis shuts down the build when we overflow 4mbs of logs:

capture


artem-zinnatullin:
Okay, another problem is that android sdk prints progress of components installation which overflows logs

artem-zinnatullin:
YIS, finally. Damn Travis, Android SDK and everything

I'll update PR title

artem-zinnatullin:
>codecov/project — No report found to compare against

That's not critical.

Issue #826, opened 2017-10-09
CI freezes.


artem-zinnatullin:
We're seeing "random" CI freezes for a long time, sometimes more often.

Latest observed logs from 2 different builds show that latest test in the output before freeze is:

```console
com.pushtorefresh.storio2.sqlite.integration.RxQueryTest > queryOneExistedObjectTableUpdate PASSED
```

So my guess is that one of the tests in `RxQueryTest` has racy deadlock. 

I'd start solving it by adding [Timeout rule](https://github.com/junit-team/junit4/wiki/Timeout-for-tests) for whole `RxQueryTest` with reasonable timeout ie 30 seconds, that way we'll figure out which test causes the problem because it'll fail the build properly.

@nikitin-da can you please try that if you have time? Otherwise I'll take care of that later :)

PR #827, opened 2017-10-09
[Not merge] Let other threads work in RxQueryTest

nikitin-da:
Part of #826

### DIFF
--- a/storio-common/src/main/java/com/pushtorefresh/storio2/internal/RxChangesBus.java
+++ b/storio-common/src/main/java/com/pushtorefresh/storio2/internal/RxChangesBus.java
@@ -15,6 +15,7 @@ class RxChangesBus<T> {
     private final Subject<T, T> rxBus = PublishSubject.<T>create().toSerialized();
 
     public void onNext(@NonNull T next) {
+        System.out.println("issue-826 RxChangesBus onNext: " + next);
         rxBus.onNext(next);
     }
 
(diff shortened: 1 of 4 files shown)


artem-zinnatullin:
Just in case there is "Raw log" button on Travis, then just search for "Failed" :)

// Regular Travis log UI freezes my computer lol

capture

nikitin-da:
Yes, I knew. It was very critical when log took 4Mb =)

artem-zinnatullin:
Almost hour of reading source code (damn I love how we wrote StorIO, so readable), changes and debugging — facepalm.jpg

`parallelWritesWithoutTransaction` is incorrect :D 

`DefaultPutResolver` does put in transaction, which means that StorIO has all rights to merge notifications occurred during transaction, @nikitin-da your super useful logs (not joking) clearly show that StorIO delivers notifications, but less than test expect, also you can see that StorIO detects notification emission during transaction in the logs and simply merges them.

Haven't looked into second test that fails, but so far I'm thinking either about deleting `parallelWritesWithoutTransaction` test or renaming and fixing assertion.

I'll continue tonight and post separate PR with fixes, again, @nikitin-da thank you for great logs!

geralt-encore:
this one can be closed, right?

nikitin-da:
@geralt-encore sure! 
Fixed via #829

PR #832, opened 2017-10-15
Fix build signing after update to Gradle 4.

artem-zinnatullin:
Auto release of 2.0.1 failed because build process didn't sign jars, sign task was skipped.

My guess is that Gradle 4.x significantly refactored build lifecycle and we can't analyze task graph at the time of deciding if we should sign jars or not.

I've checked new version locally, it runs signing, so it should be good now, I'll release v2.0.2 after merging this PR.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -200,10 +200,6 @@ allprojects {
 
 ext.VERSION_NAME = projectVersion()
 
-def isReleaseBuild() {
-    return !projectVersion().equals('development')
-}
-
 def getReleaseRepositoryUrl() {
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
Uhh, it's kinda hard to decouple release scripts and Gradle setup from Travis, it runs regular `ci.sh` before `deploy` and we're trying to sign jars right away

```console
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':storio-common:signArchives'.
> Cannot perform signing task ':storio-common:signArchives' because it has no configured signatory

```

PR #833, opened 2017-10-16
Disable debug builds for library modules.

### LINKED ISSUE #831: Don't build/test debug builds of library modules.
Looking at CI logs I've noticed that we're building and testing debug build type of library modules and then do the same for release build type.

There is no need for that since we don't have build-type specific code, so removing debug builds should speedup CI and save us some log output lines because we're slightly more than 10k atm and Travis only renders 10k lines in its UI, otherwise you have to look at raw log output.

artem-zinnatullin:
Closes #831, but doesn't solve possible Jacoco call leaks to release artifacts

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -146,6 +146,17 @@ subprojects {
                 }
             }
         }
+
+        // Ignore 'debug' buildType for library projects
+        // Because we don't have debug-specific code and we don't need to build debug variants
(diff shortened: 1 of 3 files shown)


artem-zinnatullin:
Hah @nikitin-da! No coverage loss, faster build and only 7k log lines!

PR #835, opened 2017-10-16
Remove exhaustive else from GetResolverGenerator utils

geralt-encore:


storio-common-annotations-processor/src/main/kotlin/com/pushtorefresh/storio2/common/annotations/processor/generate/Common.kt
```diff
@@ -21,6 +19,5 @@ object Common {
         DOUBLE, DOUBLE_OBJECT -> "getDouble($columnIndex)"
         STRING -> "getString($columnIndex)"
         BYTE_ARRAY -> "getBlob($columnIndex)"
-        else -> throw ProcessingException(columnMeta.element, "Can not generate GetResolver for field")
```
artem-zinnatullin: Nice, I also noticed that in build log, thanks for fixing!

PR #836, opened 2017-10-16
Do clean release build to exclude Jacoco from jar.

artem-zinnatullin:
Related to discussion in #831.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -7,14 +7,16 @@ PROJECT_DIR="$DIR"
 
 pushd "$PROJECT_DIR"
 
-# For some reason test for annotation processor are failing on a regular CI setup.
-# So we had to exclude test task for it from the main build process and execute it as a separate command.
-./gradlew clean build checkstyle -PdisablePreDex --exclude-task :storio-sqlite-annotations-processor-test:test --exclude-task :storio-content-resolver-annotations-processor-test:test
-./gradlew :storio-sqlite-annotations-processor-test:test
-./gradlew :storio-content-resolver-annotations-processor-test:test
-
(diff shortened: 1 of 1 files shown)


ci.sh
```diff
@@ -7,14 +7,16 @@ PROJECT_DIR="$DIR"
...
-# For some reason test for annotation processor are failing on a regular CI setup.
-# So we had to exclude test task for it from the main build process and execute it as a separate command.
-./gradlew clean build checkstyle -PdisablePreDex --exclude-task :storio-sqlite-annotations-processor-test:test --exclude-task :storio-content-resolver-annotations-processor-test:test
-./gradlew :storio-sqlite-annotations-processor-test:test
-./gradlew :storio-content-resolver-annotations-processor-test:test
-
 # Export "PUBLISH_RELEASE=true" to initiate release process.
-if [ "$PUBLISH_RELEASE" == "true" ]; then
+if [ "$PUBLISH_RELEASE" != "true" ]; then
+    echo "Running non-release build...".
```
artem-zinnatullin: Additionally I decided to wrap non-release build in if statement as well, otherwise it was always running repeating unnecessary build parts for publish state

artem-zinnatullin:
Uhhh

>FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':storio-common:signArchives'.
> Cannot perform signing task ':storio-common:signArchives' because it has no configured signatory

Will look later :(

artem-zinnatullin:
Ok, so I'm not sure why it doesn't work on CI, I've configured local machine very similary and used same release scripting, and it works for me locally.

My guess is that version of gpg is incompatible with one I use on mac

PR #734, opened 2016-12-12
Added generator for Table class for StorIOSQLiteProcessor. (Issue #694)

pbochenski:
Since all data for creating such class is already in annotation processors META classes it might  save some boilerplate to create such class automatically.
It generates SQL code for creating table, for additive update (only adds columns) and stores table name and column names in easy to use way.

PR #840, opened 2017-10-24
Table generator

geralt-encore:
Based on https://github.com/pushtorefresh/storio/pull/734
Thanks @pbochenski for all the heavy-lifting and sorry that I had to re-implement it instead of fixing the initial PR. I hope you are fine with it.

So any further feedback for this will be much appreciated! The only downside I see now is that it will break projects where table classes with the similar naming pattern already exist. This can be fixed by not generating table class by default, which sounds ok to me.
And we need to add some documentation about this feature as well.


artem-zinnatullin:
Looks great!

PR #841, opened 2017-10-29
Configure all signing params for automated release.

artem-zinnatullin:
Ok, I've double-checked everything locally (no daemons, no properties, etc), but on macOS. 

Turned out that even `signing.secretKeyRingFile` is required, I thought signing plugin can figure this automatically…

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -209,10 +209,6 @@ def isValidReleaseState() {
     }
 }
 
-allprojects {
-    ext.'signing.password' = System.getenv('GPG_PASSPHRASE')
-}
-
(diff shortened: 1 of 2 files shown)


artem-zinnatullin:
Travis is flucking impossible, now release just overflows logs because Travis does unconditional `./gradlew assemble` (because you know, we use Gradle, why not), then runs regular build and only then runs release…

```console
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).

The job has been terminated
```

ajakdljalrheoirh034ru09jaaoida
*angry

Issue #855, opened 2017-11-15
Add asRxCompletable to PreparedExecuteSQL


nikitin-da:
Should we rename `PreparedWriteOperation` for this purpose? Maybe `PreparedCompletableOperation`

nikitin-da:
Actually was not implemented yet

Issue #846, opened 2017-11-09
Support io.reactivex.Maybe


geralt-encore:
I'll work on this as soon as initial RxJava 2 support will be merged

nikitin-da:
It will be great!

geralt-encore:
I've started to look into Maybe implementation and not everything is quite clear.  
To be able to properly implement `MaybeOnSubscribeExecuteAsBlocking` there has to be a way to distinguish between empty and non-empty cases. Since `Result` is generic the only way to achieve it is to make `executeAsBlocking` nullable again. Is it ok or am I missing something?

geralt-encore:
Another question. What will be correct for empty list/cursor cases: to return an empty list/cursor or to just complete stream without an emission?

nikitin-da:
Dumn, my bad=( 
`preparedOperation`'s `Result` must be must be `Optional` and unwrapperd nullable value at the same time =(

Maybe we should create separate `WrappedResult` generic type, that will duplicate `Result` for `listOfObject` and `cursor`?

Otherwise, if we are ok that result of `Maybe` is `Optional` than we can create separate `MaybeOnSubscribeExecuteAsBlocking` implementations for `Optional<T>`, `List<T>` and so on.

//cc @artem-zinnatullin 

>What will be correct for empty list/cursor cases: to return an empty list/cursor or to just complete stream without an emission?

I guess we should complete stream without an emission, otherwise it will be the same `asRxSingle`

geralt-encore:
> I guess we should complete stream without an emission, otherwise it will be the same asRxSingle
Had the same thoughts and just wanted to verify it.

Not sure if I got what you were saying about `preparedOperation`'s `Result`.
Isn't actual wrapping happens on `Interceptor` level so we can return null for `Maybe` case and still be able to use `Optional` for `Single`?

nikitin-da:
Not quite so, interceptor must return value of [the same type](https://github.com/pushtorefresh/storio/blob/5850b7e450b5d58498fbc5acf23076cb055b578b/storio-sqlite/src/main/java/com/pushtorefresh/storio2/sqlite/Interceptor.java#L15) as [prepared operation uses](https://github.com/pushtorefresh/storio/blob/5850b7e450b5d58498fbc5acf23076cb055b578b/storio-sqlite/src/main/java/com/pushtorefresh/storio2/sqlite/operations/get/PreparedGet.java#L21)
[...]

PR #857, opened 2017-11-27
Support io.reactivex.Maybe

geralt-encore:
https://github.com/pushtorefresh/storio/issues/846
https://github.com/pushtorefresh/storio/issues/855

I've added Maybe support only to `PreparedGetObject`. After some thoughts, I came to a conclusion that it doesn't really make much sense to have it for other queries. Maybe is some kind of a reactive optional and in case of `List`, `Cursor` and `NumberOfResults` queries we always have a valid result for empty cases so I don't see any value in wrapping them into `Maybe`.

### DIFF
--- a/storio-common/src/main/java/com/pushtorefresh/storio2/operations/internal/MaybeOnSubscribeExecuteAsBlocking.java
+++ b/storio-common/src/main/java/com/pushtorefresh/storio2/operations/internal/MaybeOnSubscribeExecuteAsBlocking.java
@@ -1,0 +1,36 @@
+package com.pushtorefresh.storio2.operations.internal;
+
+import android.support.annotation.NonNull;
+import com.pushtorefresh.storio2.operations.PreparedMaybeOperation;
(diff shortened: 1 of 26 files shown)


storio-common/src/main/java/com/pushtorefresh/storio2/operations/internal/MaybeOnSubscribeExecuteAsBlocking.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @NonNull
+    private final PreparedMaybeOperation<Result, WrappedResult, Data> preparedOperation;
+
+    public MaybeOnSubscribeExecuteAsBlocking(@NonNull PreparedMaybeOperation<Result, WrappedResult, Data> preparedOperation) {
+        this.preparedOperation = preparedOperation;
+    }
+
+    @Override
+    public void subscribe(MaybeEmitter<Result> emitter) throws Exception {
```
nikitin-da: Nullability

PR #865, opened 2017-12-10
Add gradle versions plugin

nikitin-da:


geralt-encore:
Won't `apply plugin: 'com.github.ben-manes.versions'` inside `subprojects` closure in root `build.gradle` work instead of applying it separately in each subproject?

nikitin-da:
@geralt-encore sure, thank you

Issue #875, opened 2017-12-24
Add interop for StorIOSQLite notifications from v1 and v2


nikitin-da:
For gradually migration from v1/v2 we have to share changes between old and new StorIOSQLite.

PR #876, opened 2017-12-24
Part1: Forward notifications between StorIOSQLite v1 and v3

nikitin-da:
Part of #875

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -58,6 +58,7 @@ ext.libraries = [
         // Core libraries
         supportAnnotations                       : 'com.android.support:support-annotations:' + supportLibsVersion,
 
+        rxJava                                   : 'io.reactivex:rxjava:1.2.1',
         rxJava2                                  : 'io.reactivex.rxjava2:rxjava:2.1.6',
         rxAndroid2                               : 'io.reactivex.rxjava2:rxandroid:2.0.1',
         kotlinStdLib                             : "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version",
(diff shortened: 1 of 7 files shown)


geralt-encore:
@nikitin-da @artem-zinnatullin Do you think that it is a good idea to have it? I mean a migration to a newer version shouldn't be hard and it is a lot of code to maintain and care about.

nikitin-da:
@geralt-encore happy new year man! 🎄 🍾 🍊

nikitin-da:
>a migration to a newer version shouldn't be hard

I'm totally agree that migration of storio itself is not a big deal and can be done by single pull request.
But most likely we have to migrate to RxJava2 as well and here we can meet some problems if we use storio-rx widely. 
My current project has no separate repository that can be fast converted to storio3 but keep outer interface on rx1 by RxJavaInterop. I guess I'm not alone in this problem and many projects will have to migrate storio and rx stepwise.

>it is a lot of code to maintain and care about

Firstly I thought that some gist with notifications sharing will be enough. But lately realized that I have to migrate changes and mapping as well. It is not feet to gist and requires tests..
What about code maintaining I think to prevent induced changes in this module we may use fixed storio 3.0.0 version [here](https://github.com/pushtorefresh/storio/pull/876/files#diff-4334bc0ffc5848f5d54b8d765ee2e552R20) instead of subproject dependency.
And than it will lead to separate interop version (because otherwise interop version will outrun storio itself). 

Or maybe we should create another repository and publish it independently like it do @akarnokd.
@artem-zinnatullin any thoughts?

geralt-encore:
> @geralt-encore happy new year man! 🎄 🍾 🍊
Thanks, mate! Happy new year to you, too!

I see what problem you are trying to solve and it feels more like RxJava migration related than StorIO related. Also, it might get out of hand pretty fast. Imagine that we would release storio4 and 3 more migrations are needed. 

I like the idea of keeping it in separate repo and publishing as an independent artefact.

Issue #662, opened 2016-06-15
Any potential support for the Requery Android SQLite support library?


jedikv:
Just wondering if it's a possibility down the road to maybe have an optional extension to enable some sort of compatibility wrapper with [Requry's Android SQLite](https://github.com/requery/sqlite-android)?

Keep up the good work otherwise! It's a great library :+1:

nikitin-da:
Hi @jedikv! Thanks! 
Right now there isn't StorIO wrapper for requery directly, but it is quite simple to make it yourself. 
You just need to write own implementation `StorIOSQLite` - like [`DefaultStorIOSQLite`](https://github.com/pushtorefresh/storio/blob/master/storio-sqlite/src/main/java/com/pushtorefresh/storio/sqlite/impl/DefaultStorIOSQLite.java). 
It may be almost the same except import (`android.database.sqlite.*` --> `io.requery.android.database.sqlite.*`) and `ChangesFilter` - actually it is package private =(

@artem-zinnatullin, @karlicoss should we in the short term create separate module with Requery support?

artem-zinnatullin:
Well, that'll be nice, I believe even you will enjoy SQLite 4 in apps you
work on.

PR #907, opened 2018-11-23
Use SupportSQLite instead of the framework version

inorichi:
This PR replaces the framework SQLite version with the new SupportSQLite from AndroidX.

This new implementation allows to use more SQL implementations besides the framework one, like the [sqlite-android](https://github.com/requery/sqlite-android) database from requery, which was requested at #662.

All tests are passing and the sample app works, but I don't know if I missed something. Readme should be updated too to reflect the changes required to use the new approach with `FrameworkSQLiteOpenHelperFactory`.

I guess this PR would be included in a 4.x release :smile:


storio-sqlite/src/test/java/com/pushtorefresh/storio3/sqlite/integration/InsertTest.java
```diff
@@ -22,7 +24,7 @@ public void insertOne() {
         final User user = putUserBlocking();
 
         // why we created StorIOSQLite: nobody loves nulls
```
inorichi: This comment is now outdated :sweat_smile:

storio-sqlite/src/main/java/com/pushtorefresh/storio3/sqlite/impl/DefaultStorIOSQLite.java
```diff
@@ -424,9 +427,9 @@ public Cursor query(@NonNull Query query) {
         public long insert(@NonNull InsertQuery insertQuery, @NonNull ContentValues contentValues) {
             return sqLiteOpenHelper
                     .getWritableDatabase()
-                    .insertOrThrow(
+                    .insert(
                             insertQuery.table(),
-                            insertQuery.nullColumnHack(),
```
nikitin-da: As SupportSQLiteDatabase doesn't allow nullColumnHack, and we should remove it from InsertQuery too

storio-sqlite/src/test/java/com/pushtorefresh/storio3/sqlite/integration/InsertTest.java
nikitin-da: hah 👍

nikitin-da:
@artem-zinnatullin may we drop old SQLite support in v4?

PR #908, opened 2018-12-18
Nexus-publish-plugin 0.1.1 and providing nexusPublishing params

nikitin-da:


nikitin-da:
@artem-zinnatullin don't you know can we check case with publishing with varying IP addresses localy?

nikitin-da:
The same error, we should find the way to check it locally

artem-zinnatullin:
What's the error exactly?

nikitin-da:
Error message is `WrongNumberOfRepositories was thrown with message 'Wrong number of received repositories in state 'open'. Expected 1, received 4'.`
https://travis-ci.org/pushtorefresh/storio/builds/473768167
https://github.com/Codearte/gradle-nexus-staging-plugin/issues/76#issuecomment-451615386

nikitin-da:
uploadArchives is not supported(

